PR #5, opened 2016-10-13
[WIP] The nghttp2 glue code

jasnell:
This is a work in progress... I will update this as I go.

This starts building the process.binding('http2') API layer... which is, more or less, a shim for the nghttp2 library API. This is not yet complete.


lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+const kChunks = Symbol('chunks');
+const kProvider = Symbol('provider');
+const kPaused = Symbol('paused');
+const kResume = Symbol('resume');
+const kBeginResponse = Symbol('begin-response');
+const kEndStream = Symbol('end-stream');
+const kHasTrailers = Symbol('has-trailers');
+
+Object.defineProperty(exports, 'constants', {
+  configurable: false,
```
italoacasas, line 56: `configurable: false` is not the default value ? In that case, we need to be explicit ?

jasnell: I generally prefer to err on the side of being explicit.

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+      // code is the error code
+      this[kStreams].delete(stream.id);
+      process.nextTick(() => this.emit('stream-close', stream.id, code));
+    };
+
+    session[constants.CALLBACK_ONDATACHUNK] = (streamID, flags, data) => {
+      // streamID is the numeric stream ID the data belongs to
+      // flags is the data flags
+      // data is the Buffer containing the data
+      const stream = this[kStreams].get(id);
```
doug-wade: streamID?

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+    if (Array.isArray(value)) {
+      this[kTrailers].set(name, value.map((i) => String(i)));
+    } else {
+      this[kTrailers].set(name, String(value));
+    }
+    return this;
+  }
+
+  addTrailers(headers) {
+    var keys = Object.keys(headers);
```
doug-wade: Might

``` javascript
for (var key in headers)
  this.setTrailer(key, headers[key]);
```

be more expressive?

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+  writeHead(statusCode, headers) {
+    this.statusCode = statusCode;
+    const keys = Object.keys(headers);
+    for (var key of keys)
+      this.setHeader(key, headers[key]);
+    return this;
+  }
+
+  _write(chunk, encoding, callback) {
+    if (typeof chunk === 'string')
```
doug-wade, line 556: Do we need the encoding if a chunk is a Buffer?  Or, in other words, should encoding be an optional param, as in `end` directly below?

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+    this[kSession].sendData(this[kStream]);
+  }
+
+  end(data, encoding, callback) {
+    if (typeof encoding === 'function') {
+      callback = encoding;
+      encoding = undefined;
+    }
+    if (data) {
+      this.write(data, encoding);
```
doug-wade: this._write?

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+          stream[kRequest][kFinished] = true;
+          stream[kRequest].end();
+        }
+        process.nextTick(() => {
+          this.emit('request', stream[kRequest], stream[kResponse]);
+        });
+        break;
+      case constants.NGHTTP2_HCAT_HEADERS:
+        // ignore interstitial header blocks for now, only pay attention
+        // if the header block finishes the steam, these will be the
```
doug-wade: stream*

doug-wade (commented):
just a coupla things I noticed while I was reading through the pr -- per https://github.com/nodejs/http2/pull/6#issuecomment-253713691, I understand that this is WIP and it might be merged before they're addressed; I'm more than happy to address any or all myself once this lands.

lib/internal/http2.js
mcollina: This is part of the stream.Writable interface.

lib/internal/http2.js
mcollina: no. It is part of the stream API.

lib/internal/http2.js
```diff
@@ -0,0 +1,812 @@
...
+  }
+
+  end(data, encoding, callback) {
+    if (typeof encoding === 'function') {
+      callback = encoding;
+      encoding = undefined;
+    }
+    if (data) {
+      this.write(data, encoding);
+      super.end(callback);
```
mcollina: why are we not calling super.end(chunk, callback)?

jasnell:
@nodejs/http2 ... I'd like to get this landed even tho it's still a long ways from being done. That would allow others to help evolve this as we go.

jasnell:
Going to get this landed.

jasnell:
Landed (without commit metadata) in d2f03e98...67caa7a2

Issue #4, opened 2016-10-12
TODOs


jasnell:
- [x] Check in nghttp2 deps, initial build support
- [x] Begin src/node_http2.cc and src/node_http2.h glue code
- [x] Begin lib/internal/http2.js implementation
- [x] Basic server
- [x] Graceful server side connection shutdown
- [x] Server server timeout handling / or, how to know when the server can shut down the socket
- [ ] Are we leaking any memory?
- [ ] Tests tests test tests
- [ ] Benchmarks
- [ ] Documentation

AnshulMalik:
Hi there !

I am new here. I have been using node for a while, now I'd like to contribute :)
I think I can do some work here. Please let me know some more details of implementation and current progress.

jasnell:
@AnshulMalik ... I'm in the process of working up some design docs for this that should help.

c0b:
wonder if there are any binaries published with this http2 in core? or do I have to compile from source code if I want to test play with?

how about make another rc channel like the nightly? you don't have to publish really every night, but some kind of CI pipelines to release binaries every week would be very nice
https://nodejs.org/download/nightly/

joyeecheung:
@c0b If I understand correctly this is not shipped in any binaries at the moment, it's in its own repo to separate the issues and PRs. From https://github.com/nodejs/http2/issues/26#issuecomment-272230496 I think this is going to be shipped behind flag in Node 8(April this year).

cebrakadabra:
@jasnell does it mean, by the time all of those todos (especially tests) are checked it will be available in nodejs core? which version are you guys considering going out with it?

also @joyeecheung is this info from you the official estimation for releasing http2 in node core?

joyeecheung:
@cebrakadabra Er, no, I'm not involved in the HTTP2 implementation at the moment, this is just what I observed from https://github.com/nodejs/http2/issues/26#issuecomment-272230496

> we still plan to ship this behind a flag in Node 8.

jasnell:
No, this is just an informal list. The current plan is to try to get http/2 into 8.x as an experimental feature. When that will land is still undetermined.

c0b:
today just reading on this http://ivanjov.com/running-express-koa-and-hapi-on-http-2/ mentioned:

> Developers have created two awesome modules for working with HTTP/2 protocol, http2 and spdy.

[...]

PR #6, opened 2016-10-14
[WIP] Low level public api

doug-wade:
Per https://github.com/nodejs/node-eps/pull/38/files#diff-9192f700847790c2c21415562fc65e80R72, expose `HTTP2Session` on the low-level api (`lib/internal/http2.js` from https://github.com/nodejs/http2/issues/4).  Based on https://github.com/nodejs/http2/pull/5.

I've opened this pr super early in the implementation process to make sure I'm on the right track.  After building and requiring the module, I can instantiate an `HTTP2Session`.

```
$ ./configure && make
# stuff
$ ./node -v                                                           
v7.0.0-pre
$ ./node             
> const http2 = require('./lib/internal/http2');
undefined
> const session = new http2.HTTP2Session();
undefined
> session
Http2Session {
  localWindowSize: 65535,
  inflateDynamicTableSize: 0,
  deflateDynamicTableSize: 0,
  remoteWindowSize: 65535,
  outboundQueueSize: 0,
  lastProcStreamID: 0,
  effectiveRecvDataLength: 0,
  effectiveLocalWindowSize: 65535,
  nextStreamID: 2,
  type: 0,
  root: Http2Stream { session: [Circular], id: 0 } }
```

### DIFF
--- a/lib/internal/http2.js
+++ b/lib/internal/http2.js
@@ -1,0 +1,5 @@
+'use strict';
+
+const http2 = process.binding('http2');
+
+exports.HTTP2Session = http2.Http2Session;
--- a/node.gyp
+++ b/node.gyp
@@ -156,6 +156,7 @@
         'src/node_constants.cc',
         'src/node_contextify.cc',
         'src/node_file.cc',
+        'src/node_http2.cc',
         'src/node_http_parser.cc',
         'src/node_javascript.cc',
         'src/node_main.cc',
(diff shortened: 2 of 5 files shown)


jasnell:
I've updated 5 fairly significantly today. There's quite a bit more there. It's no where near complete, but there's a large part that is functional. Let's get that merged in the next day even tho it's not complete, and we can keep building from there.

PR #8, opened 2016-10-31
Second round of implementation upates

jasnell:
A second round of implementation updates... this is an ongoing work in progress.
This update represents a significant ongoing refactoring and filling out of the implementation and API detail.

### DIFF
--- a/configure
+++ b/configure
@@ -50,6 +50,8 @@ shared_optgroup = optparse.OptionGroup(parser, "Shared libraries",
 intl_optgroup = optparse.OptionGroup(parser, "Internationalization",
     "Flags that lets you enable i18n features in Node.js as well as which "
     "library you want to build against.")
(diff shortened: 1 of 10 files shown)


jasnell:
Will be getting this landed shortly...

AndrewBarba:
@jasnell Since this is an addition to the `http` lib is there a chance we see this released in a v6 update instead of waiting for v7 or later? Not sure how others are handling versions but our team is maintaining a strict LTS policy so we wouldn't be able to use this if it doesn't land until the next LTS.

Edit: I mean http2 as a whole. Sorry if that wasn't clear or if this PR is not the place to ask about that.

jasnell:
This is going to be a long process. It would be introduced first as an
experimental feature if the ctc decides to accept it. It should be a
semver-minor update so backporting should be possible but it's going to be
quite a while before this is ready for that

jasnell:
This pr is merged. Closing. Starting round 3

PR #9, opened 2016-11-04
Round 3

jasnell:
Round 3... continued work on the server implementation


jasnell:
@mcollina and @nodejs/http2 ... this ready to start having some attention and love paid to the Streams aspects. Specifically, the `Http2Outoing` and `Http2Incoming` classes need to be looked at. Those are the Writable and Readable Streams that channel data back and forth between the underlying nghttp2/socket and the higher level `Http2Response`, `Http2Request`, and `Http2PushResponse` APIs. Stream state management is not by strongest area so I would definitely appreciate having someone with a deeper understanding of Readable and Writing streams looking those over and filling in the missing bits. Everything you would need to look through is in `lib/internal/http2.js`.

jasnell:
I will be looking at a number of other aspects of this, including the Event Emitter flow. Right now, all of the events emitted by `Http2Session` in `lib/internal/http2.js` are emitted on nextTick, which might not be appropriate. The key consideration is that there will be multiple simultaneous streams and we have to make sure we're not blocking i/o so that the multiplexing can work appropriately.

It is likely that we will need to make `Http2Stream` a subclass of `EventEmitter` also so that we can channel events to the `Http2Stream` that are piping to and from the stream (e.g. to let the streams know if the underlying `Http2Stream` has closed or has changed state in some important way. I'll be investigating that further over the coming few days.

mcollina:
@jasnell I will do that late next week, if it's ok with you. How would you prefer me doing it? I can open issues, or directly fire PRs.

On the general side of things, I think we should avoid allocating a lot of closures (this might change) to increase speed. In all of those: https://github.com/nodejs/http2/blob/master/lib/internal/http2.js#L93-L181 we can just use top level functions, as `this` will be the EE.

jasnell:
Yeah, there is a ton of room for improvement there. I've been doing things
very incrementally so far just to make sure things work first then optimize
down to something that's better and more performant.

I will land this round 3 pr on Monday, then I'll start doing things in
smaller chunks in individual prs. Feel free to open prs or issues,
whichever you prefer!

jasnell:
Round 3 is almost ready to go. there are some timing issues on push streams that I need to resolve and there are plenty of performance issues that still need to be looked at, but it's making progress.

jasnell:
Landed

mcollina:
@jasnell how do I know I did not broke anything? Do you have any script that you are using for testing this? Maybe an end-to-end test will speed things up.

jasnell:
@mcollina... I'll be working on automated tests more this week and next. For the time being, I have been using a combination of curl with nghttp2 support (this has to be enabled specifically in the curl build), the nghttp command line client, firefox and chrome to do the testing.

jasnell:
Also, in the next round of edits I'm making improvements to how the streams API works, I want to make sure we don't accidentally duplicate effort so I'll post what I have in smaller increments in a Round 5 pr.

mcollina:
@jasnell more or less the goal should be to have backpressure on the multiplexed streams. And remove the `PassThrough`. I can give it a pass when you land some unit tests (so I have some guidance on how to use it).
Can you send me your test server?

PR #12, opened 2016-11-15
Round 4

jasnell:
Round 4 with additional improvements to performance, reliability and overall code

### DIFF
--- a/lib/internal/http2.js
+++ b/lib/internal/http2.js
@@ -8,13 +8,13 @@ const internalHttp = require('internal/http');
 const TLSServer = require('tls').Server;
 const NETServer = require('net').Server;
 const stream = require('stream');
+const BufferList = require('internal/streams/BufferList');
 const Writable = stream.Writable;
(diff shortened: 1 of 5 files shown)


jasnell:
Landed... starting round 5

lib/internal/http2.js
```diff
@@ -561,22 +553,31 @@ function mapToHeaders(map) {
...
+      const n = head.copy(buffer, offset, 0, remaining);
+      list.unshift(head.slice(remaining));
+      remaining -= n;
+      copied += n;
+    } else {
+      // head either has exactly what we need or not enough.
+      const n = head.copy(buffer, offset, 0, head.length);
+      offset += n;
+      remaining -= n;
+      copied += n;
```
mcollina, line 577: copying buffers will not be a good idea perf-wise.

jasnell:
Yeah this is the part I'm refactoring now. This was largely a temporary
measure while getting other bits working. I should have this bit
refractored by the end of today or tomorrow

PR #13, opened 2016-11-17
Round 5

jasnell:
Updates, Round 5

@mcollina ... this update turns the internal Http2Stream object into a StreamBase. Below is a basic description of how it currently works:

1. When the Http2Session object is created, it takes over ownership and control of the Socket stream. As data is read off the socket, it is immediately passed into the Http2Session object for processing by the underlying nghttp2_session object.

2. When the nghttp2_session object detects that a new HTTP/2 Stream has been created, it emits a callback. I use that callback to create and store a new Http2Stream instance.

3. Among other things, the Http2Stream instance contains one inbound and one outbound data buffer provided by the NodeBIO class (see node_crypto_bio.h/node_crypto_bio.cc). The inbound buffer (str_in_) stores the data from DATA frames *received* from the socket and processed by nghttp2_session. The oubound buffer (str_out_) stores the data written by the application code that is pending to be sent to the peer. When the nghttp2_session instance is ready to read the outbound data, it repeatedly invokes a callback that pulls it out it fixed sized chunks.

4. As I mentioned, the Http2Stream object is an extension of StreamBase. The DoWrite method pushes user-written data into str_out_, where it is held until nghttp2_session is ready to process it.

5. When user code is ready to read data from the Http2Stream, the ReadStart method calls an EmitPendingData() method that iterates through _str_in and calls EmitData for each chunk of stored data. If there is no more data pending and the HTTP/2 Stream is in the CLOSED or HALF_CLOSED_REMOTE state, then an EOF is emitted signaling that there is no more data to be read.

6. At the JavaScript layer, Http2Stream extends Duplex and maps the relevant methods to the underlying StreamBase implementation.

7. At the HTTP API layer, there are two base objects Http2Incoming and Http2Outgoing and currently two server specific extensions Http2ServerRequest and Http2ServerResponse. Every Http2Stream has both an Http2Incoming and an Http2Outgoing instance, each of which should delegate to the Http2Stream's own Duplex interface.

8. Http2ServerRequest extends Http2Incoming. Http2Incoming extends Readable.

9. Http2ServerResponse extends Http2Outgoing. Http2Outgoing extends Writable.

10. When user code reads from Http2ServerRequest (e.g. `req.on('data', (chunk) => {})`), that needs to translate down into a read on the underlying Http2Stream readable interface. When user code writes to Http2ServerResponse (e.g. `res.write('<h1>hello!</h1>');`) that needs to translate down to a write on the underlying Http2Stream writable interface.

Some caveats:

The underlying HTTP/2 Stream can be closed at any time by the nghttp_session. There can be many reasons why this happens. When the HTTP/2 Stream is closed it is analogous to a socket closing in mid-write. We need to make sure stream closures are happening gracefully. The key thing to remember, however, an HTTP/2 Stream closing does not mean that the socket has closed (it may be, but Stream lifecycle is independent of session lifecycle).

The Http2Stream object wraps the nghttp2_stream reference. When nghttp2 closes the HTTP/2 stream, the nghttp2_stream object will no longer be usable/valid, but there will still be code using the Http2Stream object. I've only put partial and temporary measures in place that would ensure that the Http2Stream object can be safely and gracefully cleaned up in a situation where the nghttp2_stream has been closed while user code is still reading/writing from the Http2Stream StreamBase interface.

The way nghttp2 reads data to be sent in a response is as follows: after the Http2Stream instance has been created, the Respond method is used to initiate a response message flow. This calls the nghttp2_submit_response API. Passed into this API is the block of response headers to send along with a reference to an nghttp2_data_provider struct. This struct defines a callback that nghttp2 will call repeatedly to fetch the data that will be chunked into the individual DATA frames. These are generally fixed sized chunks whose size is determined by nghttp2 options (usually the default max frame size). If there is no data currently to be read and the user code has not yet signaled that it is done writing (end() has not been called yet), then the callback returns NGHTTP2_ERR_DEFERRED, putting the HTTP/2 Stream into a pending state. If there is no more pending data and the user code *has* finished writing, then the ENDDATA flag is set letting nghttp2 know that there is no more data to read. If we've reached the end of the data and there are trailing headers to be sent, the callback also schedules those for transmission. Once either (a) the trailing headers are sent or (b) if there are no trailers, the final DATA frame is sent, nghttp2 will signal the peer that the stream transmission is complete, putting the HTTP/2 stream into a Local-Half-Closed state (no further writes are possible). Note that nghttp2 will synchronously call this callback repeatedly until either NGHTTP2_ERR_DEFERRED is returned or until the ENDDATA flag has been set.

When an HTTP/2 Stream has been put into deferred state by returning NGHTTP2_ERR_DEFERRED in the data read callback, nghttp2 must be explicitly instructed to resume processing of that stream data. This is done using the nghttp2_stream_resume() API.

Various methods on Http2ServerResponse (especially those that write data to the str_out_ buffer inside Http2Stream) can initiate the response process. Examples include `end()`, `write()`, `writeHead()`, and `pipe()`. Because of the typical event flow, initiating the response will typically immediately result in the stream data being put into the deferred state. This is because the first call nghttp2 makes to the data read callback generally occurs before there is any data actually buffered into the std_out_ NodeBIO buffer. What this means is that whenever outbound data is written to the Http2Stream writable interface, nghttp2 must be explicitly told to resume processing of the stream so that the buffered data can be sent.

The performance of writing the data to the underlying session and ultimately out to the socket depends a significant deal on how quickly the data written to the std_out_ buffer can be read back out by the nghttp2 callback. This depends on a significant number of factors including the total number of streams concurrently being managed by the nghttp2_session that owns the HTTP/2 stream.

If the underlying HTTP/2 Stream closes completely, or if the HTTP/2 Stream enters the Local-Half-Closed state indicating that local writes are no longer permitted to the HTTP/2 Stream, then the writable side of the Http2Stream StreamBase/Duplex should be terminated. It is possible that there might still be data stored in the str_out_ NodeBIO buffer when this occurs. We'll need to make sure that (a) writes no longer happen after this state has been reached, (b) that the buffered data is cleaned up, and (c) that a proper "aborted" event is triggered at the HTTP API layer.

When a peer *sends* data to the nghttp2_session, nghttp2 will trigger a sequence of data chunk received callbacks that include the ID of the stream, a data buffer, and a length. Currently, these are written directly to the internal str_in_ NodeBIO buffer. The nghttp2 will take care of error handling that may occur if DATA frames are received at the wrong time during a streams lifecycle so fortunately we do not have to worry about that... and once the HTTP/2 Stream has been closed, nghttp2 will stop pushing data chunks into the str_in_ buffer.

The Readable side of the Http2Stream StreamBase/Duplex has to asynchronously pull chunks of data from the str_in_ buffer and emit those chunks. Ultimately, those need to be emitted via the Http2Incoming/Http2ServerRequest Readable interface. When the HTTP/2 Stream closes and the underlying nghttp2_stream reference becomes invalid, there still may be data buffered in the str_in_ NodeBIO buffer and there still may be active consumers. We need to decide how to handle this. It is entirely likely that an async consumer just simply has not yet had enough time to completely consume the data before the underlying HTTP/2 stream is closed.

Once nghttp2 detects that a peer has completed sending all of their data frames, the nghttp2_stream will be put into a Remote-Half-Closed state. This indicates that no more data will be received into the str_in_ buffer. When this happens, once all of the pending data has been read out of str_in_ via the Http2Stream Readable interface, EOF can be emitted and the readable side of the Http2Stream can end. This is different from a full close of the nghttp2_stream because the HTTP/2 Stream can still be *written* to.

The Http2Stream object holds onto a number of resources that must be freed explicitly when the Http2Stream object is no longer being used. When the underlying nghttp2_stream becomes invalid, then the cached nghttp2_stream and nghttp2_session pointers need to be cleared. When the Http2Stream is no longer being read from or written to, and we know that we are completely done interacting with the nghttp2_stream, then the Http2Stream object needs to be explicitly destroyed.

Another consideration: HTTP/2 allows both connection and stream level flow control. As a peer is sending data frames, it is not permitted to exceed the specific Flow Control Window for either the session or the individual stream. By setting a lower number, it is possible to throttle the amount of data a peer sends. By setting the value to 0, it is possible to temporarily halt the peers ability to send any data frames. When a session initially begins, there is a default frame window size and a peer is permitted to go ahead and send that amount of data if it chooses.

One strategy for limiting the amount of bytes we have to buffer in str_in_ from DATA frames received by a peer is to set the Flow Control Window to 0, effectively turning off the flow. However, it is possible and likely that the peer will send some amount of data before the Flow Control Window can be limited. Accordingly, we have to be prepared to maintain at least *some* amount of data in the str_in_ NodeBIO buffer. We can, however, set the Flow Control Window to 0 as the default when the session is created, then use the flow control APIs within the ReadStart and ReadStop methods to start and stop the flow of data. It's not a perfect control but it would help limit the bandwidth consumption on streams that are not actively being read from. I have not yet worked this into the implementation tho.

Let's see... what else... oh, push streams... When a push stream is initiated, a new Http2Stream instance is created. The key difference is that push Http2Stream instances are not *Readable*. The Readable side of the Http2Stream StreamBase/Duplex must be closed on start up. Only the *Writable* side will be active. That said, the push Http2Stream instance will still have both a Http2ServerRequest (Http2Incoming) and Http2ServerResponse (Http2Outgoing) associated with it. The Http2ServerRequest can only be used to access the pushed request headers. The writable Http2ServerResponse side of a push Http2Stream is otherwise exactly the same as a normal Http2Stream.

The following server code can be used to create a simple functioning HTTP/2 server based on this PR:

```js
const fs = require('fs');
const http2 = require('http').HTTP2;
const options = {
  key: fs.readFileSync('/Users/james/node/main/http2/test/fixtures/keys/agent2-key.pem'),
  cert: fs.readFileSync('/Users/james/node/main/http2/test/fixtures/keys/agent2-cert.pem')
};

function handler(req, res) {
  res.writeHead(200, {'content-type': 'text/html'});
  res.end('<html><head></head><body><h1>this is some data</h2></body></html>');
}

const server = http2.createSecureServer(options, handler);
server.setTimeout(10 * 1000);
server.listen(8000);
```

Here, the `req` object is the `Http2ServerRequest` and `res` is the `Http2ServerResponse`. Each of these have a `stream` property (e.g. `req.stream` and `res.stream` that return a reference to the `Http2Stream` instance. Each should return the same instance (`req.stream === res.stream`).

Also try variations on this that use async reads and writes to and from `res` and `req`, e.g.

```js
function handler(req, res) {
  res.writeHead(200, {'content-type': 'text/html'});
  setImmediate(() => {
    res.end('<html><head></head><body><h1>this is some data</h2></body></html>');
  });
}
```

```js
const fs = require('fs');
function handler(req, res) {
  res.writeHead(200, {'content-type': 'text/html'});
  fs.createReadStream('some-file').pipe(res);
}
```

The server should be accessible from Chrome and Firefox.

For testing, I also been using the nghttp command-line client to test:

Simple GET request:
```
$ nghttp -v  https://localhost:8000
```

PUT request that sends multiple DATA frames and includes padding:
```
$ nghttp -v -H':method: PUT' -b 100 -d src/node_http2.cc https://localhost:8000
```

The nghttp2 debug output can be switched on using `./configure --debug-nghttp2`

I've been using `node --inspect` to do other tracing and profiling to ensure no memory leaks, etc.

### DIFF
--- a/lib/internal/http2.js
+++ b/lib/internal/http2.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const http2 = process.binding('http2');
+const timers = require('timers');
 const util = require('util');
 const Buffer = require('buffer').Buffer;
 const EventEmitter = require('events');
@@ -8,32 +9,31 @@ const internalHttp = require('internal/http');
 const TLSServer = require('tls').Server;
 const NETServer = require('net').Server;
 const stream = require('stream');
-const BufferList = require('internal/streams/BufferList');
+const WriteWrap = process.binding('stream_wrap').WriteWrap;
+const ShutdownWrap = process.binding('stream_wrap').ShutdownWrap;
+const uv = process.binding('uv');
 const Writable = stream.Writable;
-const PassThrough = stream.PassThrough;
+const Readable = stream.Readable;
+const Duplex = stream.Duplex;
 const constants = http2.constants;
 const utcDate = internalHttp.utcDate;
 
+const kStatusCode = Symbol('status-code');
 const kOptions = Symbol('options');
 const kStream = Symbol('stream');
 const kHeaders = Symbol('headers');
 const kHeadersSent = Symbol('headers-sent');
-const kTrailersSent = Symbol('trailers-sent');
 const kSession = Symbol('session');
 const kRequest = Symbol('request');
 const kResponse = Symbol('response');
 const kFinished = Symbol('finished');
 const kTrailers = Symbol('trailers');
-const kChunks = Symbol('chunks');
-const kProvider = Symbol('provider');
 const kResume = Symbol('resume');
 const kBeginSend = Symbol('begin-send');
-const kEndStream = Symbol('end-stream');
-const kHasTrailers = Symbol('has-trailers');
 const kInFlight = Symbol('in-flight');
-const kExpectContinue = Symbol('expect-continue');
-const kResponseFlags = Symbol('response-flags');
-const kResponseFlag_SendDate = 0x1;
+const kSendDate = Symbol('send-date');
+const kDetached = Symbol('detached');
+const kNoBody = Symbol('nobody')
 
 const kDefaultSocketTimeout = 2 * 60 * 1000;
 const kRenegTest = /TLS session renegotiation disabled for this socket/;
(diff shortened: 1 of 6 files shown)


lib/internal/http2.js
```diff
@@ -48,7 +46,139 @@ Object.defineProperty(exports, 'constants', {
...
 // it has to inherit from EventEmitter.
 util.inherits(http2.Http2Session, EventEmitter);
+util.inherits(http2.Http2Stream, Duplex);
 
+function maybeDestroyStream(stream) {
+  if ((!stream.readable &&
+       !stream.writable &&
+       !stream._writableState.length)) {
+    stream.destroy();
+  }
```
mcollina, line 80: the best check here is checking if `destroy` exists. There is no guarantee, as its not part of the API.
Ideally we can do `stream.destroy(err)`.

lib/internal/http2.js
```diff
@@ -48,7 +46,139 @@ Object.defineProperty(exports, 'constants', {
...
+    err = createWriteReq(req, this, data, enc);
+  }
+
+  if (err) {
+    throw util._errnoException(err, 'write', req.error);
+  }
+
+  this._bytesDispatched += req.bytes;
+
+  cb();
```
mcollina: should this `cb` be called in `req.oncomplete`? Otherwise we will not be handling backpressure?

mcollina:
> One strategy for limiting the amount of bytes we have to buffer in str_in_ from DATA frames received by a peer is to set the Flow Control Window to 0, effectively turning off the flow. However, it is possible and likely that the peer will send some amount of data before the Flow Control Window can be limited. Accordingly, we have to be prepared to maintain at least some amount of data in the str_in_ NodeBIO buffer. We can, however, set the Flow Control Window to 0 as the default when the session is created, then use the flow control APIs within the ReadStart and ReadStop methods to start and stop the flow of data. It's not a perfect control but it would help limit the bandwidth consumption on streams that are not actively being read from. I have not yet worked this into the implementation tho.

I think we can start with a Flow Control Window of 32KB. This means we can fill in all the buffers of all streams when they are allocated. We have a 16KB buffer on the socket and a 16KB buffer on the Http2Session.
However, I think you are right about maintaining some data in str_in_ NodeBIO buffer: we should do it.

lib/internal/http2.js
jasnell: Probably :-)

lib/internal/http2.js
jasnell: Which `destroy`? On `stream` or on `_writableState`? The `stream.destroy()` method is exposed by `Http2Stream` in order to free up the underlying resources.

Some background: There are three key states that need to be tracked: (1) Is the Http2Stream attached to the underlying Http2Session (e.g. that is, is the internal nghttp2_stream handle still valid), (2) Is the Http2Stream Duplex interface still Writable, and (3) Is the Http2Stream Duplex still Readable. The Http2Stream object cannot be disposed of and garbage collected until all three of those states are false. Because of async reads/writes on the Duplex interface, it's possible for the Http2Stream to become detached from the underlying Http2Session. Ideally, when that happens both the Readable and Writable sides will close and any internally buffered data would be dropped and additional writes will fail but I do not have that completely wired up yet.

The `maybeDestroyStream(stream)` check is largely a temporary stopgap to determine when the Writable and Readable sides are closed so that we can allow the Http2Stream to be garbage collected.

mcollina: that's perfect. Can you put the above description as a comment on top of the function? it's definitely not clear :D. It seems the usual pattern "destroy if there is a destroy function", which I have written thousands of time.

lib/internal/http2.js
```diff
@@ -48,7 +46,139 @@ Object.defineProperty(exports, 'constants', {
...
+    //maybeDestroy(self);
+  }
+};
+
+http2.Http2Stream.prototype.end = function(chunk, encoding, callback) {
+  Duplex.prototype.end.call(this, chunk, encoding, callback);
+  var req = new ShutdownWrap();
+  req.oncomplete = (status, handle, req) => {
+    process.nextTick(() => maybeDestroyStream(this));
+  };
```
mcollina: we can probably avoid this closure, by adding a reference to the http2stream to ShutdownWrap, and use a top level function.

lib/internal/http2.js
```diff
@@ -48,7 +46,139 @@ Object.defineProperty(exports, 'constants', {
...
+    case 'utf-16le':
+      return handle.writeUcs2String(req, data);
+
+    default:
+      return handle.writeBuffer(req, Buffer.from(data, encoding));
+  }
+}
+
+http2.Http2Stream.prototype._writev = function(chunks, cb) {
+  this._writeGeneric(true, chunks, '', cb);
```
mcollina: I would recommend against this type of polymorphism. This is an hot path, and monorphic functions are always faster.
Let me know when you want me to do an optimization pass.

mcollina:
Anyway, LGTM.

Should we set up 1-2 hours to work a bit on this in Austin?

jasnell:
Yes! Absolutely let's work on it in Austin. I'll still be working on it in the meantime also. The main thing I'm hitting up against right now is how to best have `Http2ServerRequest`/`Http2Incoming` effectively proxy the Readable side of the `Http2Stream` Duplex.

tatsuhiro-t:
I have to say that nghttp2_stream life time is very short according to the official manual:
https://nghttp2.org/documentation/nghttp2_session_find_stream.html

Currently, it looks like nghttp2_stream exists until it is closed.
nghttp2_session_find_stream does only one hash lookup, and relatively cheap, so it might be better to get it every time if you want it.  It eliminates dangling pointer which is very hard to track down in some cases.

jasnell:
@tatsuhiro-t ... yes, I've noticed that as well and thank you for bringing it up. At the very least that should make it somewhat less fragile. Good call. :-)

lib/internal/http2.js
jasnell: Will do. Let me finish up this round of refactoring tonight and get this PR landed. After that I'll leave this part of the code alone and let you do your thing.

jasnell:
@mcollina ... ok done! just landed this PR into the master. I'll go work on a few other things for a few days and let you do your thing! :-)

Before landing this I address the additional comments included here including @tatsuhiro-t's suggestion not the cache the nghttp2_stream pointer.

jasnell:
As a point of reference, using the h2load tool to benchmark using a simple server instance that writes a small amount of data back, the numbers I'm getting currently when running h2load against a localhost server:

```
$ h2load -n 100000 -c 10 https://localhost:8000
starting benchmark...
spawning thread #0: 10 total client(s). 100000 total requests
TLS Protocol: TLSv1.2
Cipher: ECDHE-RSA-AES128-GCM-SHA256
Server Temp Key: ECDH P-256 256 bits
Application protocol: h2
progress: 10% done
progress: 20% done
progress: 30% done
progress: 40% done
progress: 50% done
progress: 60% done
progress: 70% done
progress: 80% done
progress: 90% done
progress: 100% done

finished in 7.34s, 13626.23 req/s, 1.96MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 14.40MB (15102100) total, 294.84KB (301920) headers (space savings 95.28%), 12.40MB (13000000) data
                     min         max         mean         sd        +/- sd
time for request:      117us     29.62ms       728us       769us    99.14%
time for connect:     6.83ms     15.07ms     11.44ms      2.94ms    60.00%
time to 1st byte:    14.36ms     15.99ms     15.13ms       496us    50.00%
req/s           :    1362.68     1369.14     1365.62        2.18    70.00%
```

Running the same test on the same server but load balanced using cluster, the numbers are:

```
$ h2load -n 100000 -c 10 https://localhost:8000
starting benchmark...
spawning thread #0: 10 total client(s). 100000 total requests
TLS Protocol: TLSv1.2
Cipher: ECDHE-RSA-AES128-GCM-SHA256
Server Temp Key: ECDH P-256 256 bits
Application protocol: h2
progress: 10% done
progress: 20% done
progress: 30% done
progress: 40% done
progress: 50% done
progress: 60% done
progress: 70% done
progress: 80% done
progress: 90% done
progress: 100% done

finished in 3.40s, 29373.30 req/s, 4.23MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 14.40MB (15101348) total, 294.11KB (301168) headers (space savings 95.29%), 12.40MB (13000000) data
                     min         max         mean         sd        +/- sd
time for request:      102us     49.93ms       322us       439us    98.88%
time for connect:    13.73ms     16.87ms     14.93ms      1.19ms    60.00%
time to 1st byte:    19.77ms     24.60ms     21.20ms      1.61ms    70.00%
req/s           :    2937.59     3172.93     3047.91       86.74    60.00%
```

jasnell:
btw, @tatsuhiro-t .. thank you so very much for taking a look at this. I appreciate the work you've done on nghttp2 and I really do consider it a beautiful piece of code. Thank you for giving us such a solid base to build on.

tatsuhiro-t:
@jasnell Thank you.  That really encourages us.  Feel free to ping me about nghttp2 stuff.  I don't know about nodejs internals, but I can develop nghttp2 a little.

PR #14, opened 2016-11-22
Round6

jasnell:
Next round of updates... this one is pretty significant.

the `Http2Stream` and `Http2Session` objects exposed by `process.binding('http2')` are now `_handle`'s of externally defined `Http2Stream` and `Http2Session` JavaScript objects. FreeList is used to limit creation of new objects and reuse in the same way we reuse http_parser instances.

There are three fundamental performance bottlenecks in this:

* How fast we can read/write data from the socket and pass data to/from the nghttp session instance
* How fast we can process the callbacks invoked by nghttp2, some of which end up requiring us to cross the JS barrier to call into user code
* How fast we can manage the flow of data via the Streams API back into nghttp2

The first two have been fairly well optimized but more can definitely be done. The Streams API implementation definitely needs a lot more work.

### DIFF
--- a/lib/_tls_wrap.js
+++ b/lib/_tls_wrap.js
@@ -346,6 +346,10 @@ TLSSocket.prototype.disableRenegotiation = function disableRenegotiation() {
   this[kDisableRenegotiation] = true;
 };
 
+TLSSocket.prototype.disableRenegotiation = function disableRenegotiation() {
+  this[kDisableRenegotiation] = true;
+};
+
 TLSSocket.prototype._wrapHandle = function(wrap) {
   var res;
   var handle;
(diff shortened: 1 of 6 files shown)


jasnell:
Looking at some benchmark numbers. Using a simple test clustered server that returns about 160 bytes of data, comparing this implementation with the http2 module out on npm currently. The benchmark issues 100000 requests, spread across 8 threads  (12500 requests per thread).

Method|req/s|completion time
------|-----|----
http2 module|484.72|206.30s
nghttp2 impl|18526.92|5.40s

*slight* performance difference ;-)

In comparison, running the same test using the HTTP/1 implementation in core:

Method|req/s|completion time
------|-----|----
http1|19712.39|5.07s

mightyiam:
:open_mouth:

jasnell:
Updated in master now... starting round 7

PR #21, opened 2016-12-05
more perf updates

mcollina:
cc @jasnell

### DIFF
--- a/lib/internal/http2.js
+++ b/lib/internal/http2.js
@@ -1307,6 +1307,11 @@ function initializeTLSOptions(options) {
   return options;
 }
 
+function onErrorSecureServerSession(err, conn) {
+  if (!this.emit('clientError', err, conn))
+    conn.destroy(err);
+}
+
 class Http2SecureServerSession extends TLSServer {
   constructor(options, requestListener) {
     super(initializeTLSOptions(options), connectionListener);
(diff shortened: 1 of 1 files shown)


lib/internal/http2.js
```diff
@@ -1605,20 +1606,7 @@ class Http2ClientRequest extends Http2Outgoing {
...
-              if (Array.isArray(value) && value.length > 0) {
-                for (const item of value)
-                  this.stream.addTrailer(key, String(item));
-              } else {
-                this.stream.addTrailer(key, String(value));
-              }
-            }
-          }
-        });
+        this.stream.once('handle', kBeginSendHandle(this))
```
targos: Do you know why perf is better with this change?

lib/internal/http2.js
```diff
@@ -1629,6 +1617,23 @@ class Http2ClientRequest extends Http2Outgoing {
   }
 }
 
+function kBeginSendHandle(that) {
```
targos: don't we use the prefix `k` for constants?

lib/internal/http2.js
mcollina: the function declaration is hoisted. It happens even if the if statement is false.

lib/internal/http2.js
mcollina: I just called it in a way that will show where it is used. I'll drop the k.

mcollina:
@targos check it out now. I've removed all the closure creation, and the for loop can be optimized by V8 (short-lived closures are currently not optimized).

targos:
Nice! LGTM

jasnell (approved):

jasnell:
Landed via https://github.com/nodejs/http2/pull/23

PR #23, opened 2016-12-05
No closures

mcollina:
includes #21.

I have used bind in some cases where perf was not needed and we needed to pass some parameters to the function. `bind` is as fast as creating a closure in the latest V8, but it is not hoisted and so we avoid it altogether for the happy-path.

### DIFF
--- a/lib/internal/http2.js
+++ b/lib/internal/http2.js
@@ -628,8 +628,8 @@ class Http2Incoming extends Readable {
     super(initHttp2IncomingOptions(options));
     if (!(stream instanceof Http2Stream))
       throw new TypeError('stream argument must be an Http2Stream instance');
-    if (!(headers instanceof Map))
-      throw new TypeError('headers argument must be a Map');
+    if (typeof headers !== 'object')
+      throw new TypeError('headers argument must be an object');
     this[kStream] = stream;
     this[kHeaders] = headers;
     this[kFinished] = false;
@@ -679,19 +679,19 @@ class Http2ServerRequest extends Http2Incoming {
   }
 
   get method() {
-    return this.headers.get(constants.HTTP2_HEADER_METHOD);
+    return this.headers[constants.HTTP2_HEADER_METHOD];
   }
 
   get authority() {
-    return this.headers.get(constants.HTTP2_HEADER_AUTHORITY);
+    return this.headers[constants.HTTP2_HEADER_AUTHORITY];
   }
 
   get scheme() {
-    return this.headers.get(constants.HTTP2_HEADER_SCHEME);
+    return this.headers[constants.HTTP2_HEADER_SCHEME];
   }
 
   get url() {
-    return this.headers.get(constants.HTTP2_HEADER_PATH);
+    return this.headers[constants.HTTP2_HEADER_PATH];
   }
 }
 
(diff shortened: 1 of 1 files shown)


jasnell (approved):

jasnell:
Want to close #21 since this includes it?

lib/internal/http2.js
```diff
@@ -608,9 +601,7 @@ function incomingOnRead(nread, buffer) {
       }
     }
     return;
-  }
-
-  if (nread === 0) {
+  } else if if (nread === 0) {
```
evanlucas: Should this just be } else if (nread === 0) {?

mcollina: Oh, I did a mistake here. Those changes were not supposed to go through. not sure why it's still valid JS.

mcollina:
@jasnell yes.

@evanlucas adresses.

lib/internal/http2.js
mightyiam: Sorry. I had to tweet this.
https://twitter.com/mightyiam/status/805822151462895616
Thing is, I don't know, either. Is it valid?

lib/internal/http2.js
```diff
@@ -1629,6 +1611,22 @@ class Http2ClientRequest extends Http2Outgoing {
   }
 }
 
+function addTrailers () {
+  const request = this[kRequest];
+  if (request[kTrailers] instanceof Map) {
```
evanlucas: should we use the internal util for this? Is there a chance that the Map was created in a different context?

lib/internal/http2.js
```diff
@@ -1629,6 +1611,22 @@ class Http2ClientRequest extends Http2Outgoing {
...
+    for (const v of request[kTrailers]) {
+      const key = String(v[0]);
+      const value = v[1];
+      if (Array.isArray(value) && value.length > 0) {
```
evanlucas, line 1622: Doesn't for of deopt for arrays?

lib/internal/http2.js
jasnell: @evanlucas ... no, the Map here should always be created in the same context. We could likely just have this be a truthy check to be honest.

lib/internal/http2.js
jasnell: hmm... I believe it does... (or at least did... I don't know if it still does)

mcollina:
@evanlucas @jasnell I moved away from Map() and converted it back to `{}`. It should solve most of the issues. Unfortunately for-of loops are really slow :(.

evanlucas:
@mcollina won't moving back to an object prevent setting headers like `__proto__`?

mcollina:
@evanlucas this is the behavior that we have on `require('http')`, and it is faster than `Map` for iterating over all the pairs in the map. 

https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L365

There is PR to change that, see https://github.com/nodejs/node/pull/6102.

targos:
I suppose http2 support won't land before v8.0.0? By that time, we will have a much more recent version of V8 in which the for-of loop could be fast enough.
I think we should'nt make those engine-related micro optimizations too early.

mightyiam:
I like what @targos is saying. Optimise as required, when required.

mcollina:
I strongly disagree. There are both perf and compatibility issue with http1 -> see the discussion there.
As far as I understand, there is no magical "for of" optimization landing anytime soon in V8.

This particular code makes no sense if it is not as fast as possible.

jasnell:
@mcollina ... switching away from `Map` is fine but I think using a prototype-less object (`Object.create(null)`) would be better. We wanted to make such a change in the http/1 stuff but haven't been able to due to backwards compatibility issues that we simply do not have in this case. Using the prototype-less object allows us to not have to deal with issues like `__proto__`

jasnell:
@targos: the goal is to get this into 8.x as close to 8.0.0 as possible as experimental.

mcollina:
@jasnell should be good now. Please review.

jasnell (approved):

jasnell:
@mcollina ... I'm going to get this landed so I can start building off of it.

jasnell:
Landed!

PR #25, opened 2016-12-07
Removed stream accessor

mcollina:
It was coming out hot in the flamegraph, not a noticeable improvement,
but happy to have it anyway.


jasnell:
hmm... there's a tradeoff here. The primary motivation for using symbols here is to hide the state as much as possible and to discourage monkeypatching by the ecosystem. I'd really like to avoid this change if we can.

mcollina:
I agree, but on `stream` we are essentially exposing it anyway.

A different solution is that all internal access pass through `const stream = this[kStream]` rather than `this.stream`.

jasnell:
well, there's a getter for stream but not a setter...

PR #28, opened 2016-12-09
http2: prefer kStream to .stream accessor to speed things up.

mcollina:
superseded #25. The stream accessor is not a hot function anymore.


jasnell (approved):

jasnell:
Landed in master. Also, I squashed things down in master and force pushed to clean up

Issue #26, opened 2016-12-07
HTTP2 meeting


mcollina:
We should have a quick call to discuss how to progress this better.

How about the 20th Dec at 5pm GMT? Does it work for everyone?

Join: https://hangouts.google.com/call/t7zou5rcnnf5begg7dqqicjbcie
Watch: https://www.youtube.com/watch?v=uQtKyX0eB2A

mcollina:
Is this on today?

jasnell:
Oy... unfortunately I completely forgot about this one and have a schedule conflict at 5pm GMT :-( ... would it be possible to move it to Thursday?

mcollina:
There are no issues on my side. Can we do it a bit earlier? Maybe at 17 or
18 GMT?

I'll send you an invite.
Il giorno mar 20 dic 2016 alle 16:27 James M Snell <notifications@github.com>
ha scritto:

mcollina:
@jasnell ping, should we do this call over this week?

jasnell:
Yes, a call would be great. What days/times would be best? Want to put together a doodle? I can pretty much make any time work this week

mcollina:
Let's a do a doodle, as others might want to join.

evanlucas:
yay, I would be interested in joining if y'all don't mind. Thanks!

italoacasas:
Can you guys make sure of recording this? I'm on vacations but I still have pending the "list of tests that needs to be added"...

mcollina:
Hey folks, here is a doodle: http://doodle.com/poll/ptudtfhtgpismfzd

targos:
@mcollina this link is 404.

mcollina:
@targos updated http://doodle.com/poll/ptudtfhtgpismfzd.

mcollina:
I've closed the doodle for tomorrow, Thursday 12th at 17:00 GMT.

Should I set up the hangout and the recording?

jasnell:
yes please!

williamkapke:
Added to the Calendar!
https://calendar.google.com/calendar/embed?src=kap.co_i17i575te0aes6kaanfjr2e4hs%40group.calendar.google.com

PS: there are several meeting following this that plan to use the live channel- so don't go over 1hr!

mcollina:
I've added the links to join or watch at the top.

jasnell:
[...]

PR #31, opened 2017-01-08
Round 8

jasnell:
@mcollina @nodejs/http2 ... took a few steps back and refactored the internals significantly. Yields a fairly significant performance boost. Will be continuing to evolve this over the next couple of weeks


mcollina:
I'm starting to think we should not inherit from `Readable` and  `Writable`, but just from `Stream` (if at all) for the incoming request and response. Firstly,`Readable`  and `Writable`  do a lot of buffering, which we are already doing inside the `Duplex` stream. Secondly, there are plenty of calls to `resume()`, which is a code smell to me. This is "streams 1" API.

`require('http')` use `Stream` and streams 1 API.

jasnell:
I'm all for whatever would make it easier and more performant. Most of the performance limitation in this right now is due to the Streams interface along with the need to make the high level API similar to the existing http/1 API. For instance, I'd really prefer getting rid of the separate `Http2Request` and `Http2Response` objects entirely and just use the `Http2Stream` object for both, but doing so would require a more significant high level API change.

mcollina:
@jasnell would it be feasible to have a lower-level Http2Stream that can be sufficient to build apps only on HTTP2, and then provide a compat layer to match `require('http')`? If it's what you are talking about, go for it.

jasnell:
It certainly would make things easier. The trick is designing that lower level interface in a way that would allow the higher level interface to still be performant. Also, even with the lower level interface, we need to decide how closely we want to bind it to the HTTP specific semantics vs. just exposing the Frame/Stream/Session model. The choice there is going to decide what kinds of performance optimizations we can make. For example, the decision about whether or not to attach a `DataProvider` to a response needs to be made when we initiate the response and can be based on a number of factors (what status code are we using? do we know if there is data pending to be written? is the Stream interface on the JavaScript side been ended yet?). If we always say there is a `DataProvider`, then we end up with a situation where there is always at least one potentially empty DATA frame sent on every stream, which is wasteful. The point is, the performance profile is going to depend entirely on how closely bound to the HTTP semantics we need this to be, and the more we push the HTTP semantics into the JS side, the fewer HTTP-specific performance optimizations we can make in the native code.

mcollina:
@jasnell I think this is a great topic for #26 .

jasnell:
Landed in master

PR #34, opened 2017-01-13
Round 9

jasnell:
/cc @mcollina 

Here we go....


test/parallel/test-http2-server-startup.js
```diff
@@ -6,11 +6,11 @@
```
thefourtheye, line 6: File permissions are changed.

mcollina (approved):
LGTM

lib/internal/http2/util.js
```diff
@@ -0,0 +1,109 @@
...
+  isIllegalConnectionSpecificHeader,
+  llistToHeaders,
+  mapToHeaders
+};
+
+
+
+/*
+onStream stuff
+  var request =
```
YurySolovyov, line 73: is that comment really needed?

jasnell: This is just temporary. Keeping around some of the refactored out code so it's not lost as we shift a few other bits around.

italoacasas:
@jasnell do you think make sense to assert that the fallowing function are there ?
```js
- createServer()
- createSecureServer()
- createServerSession()
- createClientSession()
```

jasnell:
Yeah, I'm thinking we should assert the presence of all the primary APIs. The tests here only barely scratch the surface.

jasnell:
Quite a few additional updates here. For the most part this is pretty solid. Push is now enabled.

@mcollina: There are limits in here. In a load benchmark throwing 1,000,000 requests at once, most requests fail and eventually the server will hit a segfault attempting to free some memory. I haven't yet tracked down where that segfault is happening.

jasnell:
@mcollina ... I'm going to get this round landed. The current implementation has no apparently leaks and the segfault appears to be closed (at least in my most recent tests).

jasnell:
Landed... will start working on round10

PR #35, opened 2017-01-29
A couple of refactoring suggestions

addaleax:
Just started taking a (so far not very exhaustive) look at the sources here and got a couple of suggestions (and at least one minor bugfix).

A lot of this is optional so really, feel free to leave commits out you don’t like.

### DIFF
--- a/doc/api/tls.md
+++ b/doc/api/tls.md
@@ -543,7 +543,7 @@ Returns the reason why the peer's certificate was not been verified. This
 property is set only when `tlsSocket.authorized === false`.
 
 ### tlsSocket.disableRenegotiation()
-<!~- YAML
+<!-- YAML
 added: REPLACEME
 -->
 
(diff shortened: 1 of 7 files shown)


mcollina (approved):
LGTM

jasnell (approved):

addaleax:
Are there any rules on landing PRs here? Should I or one of you? Should I run CI (tests aren’t passing on master here so I’m very “ummm…” on that)? Should we wait 48/72 hours?

jasnell:
Keeping it pretty loose right now. Go ahead and land if you'd like! In another week or two we should look at firming up the process.

addaleax:
Ok. I assume at some point you’ll submit a PR to the main repo where all the work here is squashed together, so it doesn’t matter whether I do that during landing?

jasnell:
Yep, I'll be refactoring the commits once this is ready to land in master as experimental. Squashing these before landing would be beneficial to help keep things organized but it's not absolutely necessary

addaleax:
Landed in 6deff9ecc229...4af0278e4ec8 – I’ve just squashed the Freelist-related commits together, I think that makes sense.

PR #36, opened 2017-01-31
Round: 10: performance improvements

jasnell:
The start of round 10....

### DIFF
--- a/src/node-http2-core-inl.h
+++ b/src/node-http2-core-inl.h
@@ -122,7 +122,7 @@ struct DataChunksFreelistTraits : public DefaultFreelistTraits {
   template <typename T>
   static T* Alloc() {
     nghttp2_data_chunks_t* chunks = Calloc<nghttp2_data_chunks_t>(1);
-    chunks->buf.AllocateSufficientStorage(kSimultaneousBufferCount);
+    chunks->buf.AllocateSufficientStorage(MAX_BUFFER_COUNT);
     return chunks;
   }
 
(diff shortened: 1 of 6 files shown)


src/node-http2-core-inl.h
```diff
@@ -195,6 +194,13 @@ inline bool nghttp2_session_find_stream(
   }
 }
 
+inline void nghttp2_set_callbacks_allocate_send_buf(
+    node_nghttp2_session_callbacks* callbacks,
+    nghttp2_allocate_send_buf_cb cb) {
+  assert(callbacks != nullptr);
+  callbacks->allocate_send_buf = cb;
+}
```
addaleax, line 202: I find these functions a bit curious… what are the reasons not to inline them at their call sites, or at least make them methods on `node_nghttp2_session_callbacks`?

src/stream_base.h
```diff
@@ -93,6 +93,17 @@ class WriteWrap: public ReqWrap<uv_write_t>,
```
addaleax, line 93: What’s up with these file access mode changes? :D

jasnell: Windows and I keep forgetting to tell git to ignore them.

src/node-http2-core-inl.h
jasnell: started this by following the pattern nghttp uses but I've been wanting to refactor it to simplify.

jasnell:
Landed

mcollina:
🎉

jasnell:
Will start on round 11 next week as time allows. Also @addaleax has offered to do some digging into the code also... I'm quite certain she'll be able to find many ways of improving it further :-)

PR #37, opened 2017-02-01
http2: enable the compatibility mode

mcollina:
cc @jasnell 

There are a couple of functions I could not find, do you still have the code around?

I need to still work on this anyway.


mcollina:
@jasnell from my tests the performance of this is almost on par with Http2Stream.
It's far from complete and it needs more testing to be 100% compliant with the Stream API.

jasnell:
excellent. I do have another update pending. Are you basing this off master or my round 10 pr?

jasnell:
any code removed should be either in the compat or util js files commented out.

mcollina:
this is on master, but those missing things are not to be found:

`isPseudoHeader`
`isIllegalConnectionSpecificHeader`
`llistToHeaders`

mcollina:
I can rebase on top of round10 before landing.

jasnell:
Each of those methods is here: https://github.com/nodejs/http2/blob/master/lib/internal/http2/util.js

You likely don't need to worry about `llistToHeaders` in the compat layer. 

There is a fix I still need to do here to make sure that `:' prefixed headers appear first in the list passed down to the binding layer.

mcollina:
@jasnell those method does not work with the current code, as the headers in `stream.respond()` are an object, while they produce an array.

jasnell:
Yeah, the methods have changed a bit. I'll work up an example of what's needed on the input shortly

mcollina:
> Yeah, the methods have changed a bit. I'll work up an example of what's needed on the input shortly

@jasnell ping :)

jasnell:
oh sorry, lol..

Here is a minimal server:

```js
'use strict';

const http2 = require('http2');
const util = require('util');
const constants = http2.constants;

const options = {};

const server = http2.createServer(options, (stream, headers) => {
  stream.respond({
    [constants.HTTP2_HEADER_STATUS]: 200,
    'foo': 'bar'
  });
  stream.end('ok');
});

server.listen(8000);
```
The first argument to the `stream.respond()` method is an object that will be translated into the appropriate array structure internally.

mcollina:
this is good to go now, minus #40 to be solved.

mcollina:
@jasnell this is ready to go, may I merge?

jasnell (approved):

mcollina:
I've fixed a broken test and linted everything as well, hope you don't mind.

jasnell:
Woot! Lgtm

PR #39, opened 2017-02-06
Fix links

sericaia:
Links should be markdown and not html


mcollina:
hi @sericaia this repo is only for http2 dev, so probably you should send this PR to https://github.com/nodejs/node

jasnell:
The links are there for the generated html on the nodejs.org website, where the pages are converted into *.html. 

Going to go ahead and close this PR. Thank you for the contribution, however. It really is appreciate that you took the time!

PR #41, opened 2017-02-18
C++ refactor

addaleax:
This is mostly just moving things from C-style language to more idomatic C++ where I thought it made sense

CI on this PR: https://ci.nodejs.org/job/node-test-commit/8008/
CI on http2/master for comparison: https://ci.nodejs.org/job/node-test-commit/8009/

### DIFF
--- a/src/node-http2-core-inl.h
+++ b/src/node-http2-core-inl.h
@@ -183,15 +183,13 @@ inline bool Nghttp2Session::HasStream(int32_t id) {
   return s != streams.end();
 }
 
-inline bool Nghttp2Session::FindStream(
-    int32_t id,
-    std::shared_ptr<nghttp2_stream_t>* stream_handle) {
+inline std::shared_ptr<nghttp2_stream_t> Nghttp2Session::FindStream(
+    int32_t id) {
   auto s = streams.find(id);
   if (s != streams.end()) {
-    *stream_handle = s->second;
-    return true;
+    return s->second;
   } else {
-    return false;
+    return std::shared_ptr<nghttp2_stream_t>(nullptr);
   }
 }
 
@@ -208,7 +206,8 @@ ssize_t Nghttp2Session::OnStreamRead(nghttp2_session* session,
                                      void* user_data) {
   Nghttp2Session* handle = static_cast<Nghttp2Session*>(user_data);
   std::shared_ptr<nghttp2_stream_t> stream_handle;
-  assert(handle->FindStream(stream_id, &stream_handle));
+  stream_handle = handle->FindStream(stream_id);
+  assert(stream_handle);
 
   size_t remaining = length;
   size_t offset = 0;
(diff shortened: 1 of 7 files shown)


src/node-http2-core-inl.h
```diff
@@ -245,6 +245,8 @@ ssize_t Nghttp2Session::OnStreamRead(nghttp2_session* session,
   int writable = stream_handle->queue_head_ != nullptr ||
                  nghttp2_stream_writable(stream_handle);
   if (offset == 0 && writable && stream_handle->queue_head_ == nullptr) {
+    /* TODO(addaleax): ask @jasnell what the correct semantics are...
+       this is dead code right now */
```
addaleax: @jasnell ;)

src/node-http2-core.cc
```diff
@@ -0,0 +1,279 @@
...
+    delete head;
+  }
+
+ end:
+  int writable = stream_handle->queue_head_ != nullptr ||
+                 stream_handle->IsWritable();
+  if (offset == 0 && writable && stream_handle->queue_head_ == nullptr) {
+    /* TODO(addaleax): ask @jasnell what the correct semantics are...
+       this is dead code right now */
+    return NGHTTP2_ERR_DEFERRED;
```
jasnell: @addaleax NGHTTP2_ERR_DEFERRED needs to be returned when the stream writable side has not been closed but there is no data currently available to send. What this return code does is tell nghttp2 to take this stream temporarily out of the send queue until data is available (which will happen on the next write).

jasnell: More specifically, when told to do so, for each stream in the send queue, nghttp will repeatedly call the read function to grab chunks of data. As long as there is data available, it will continue call. If the function returns `NGHTTP2_ERR_DEFERRED`, the stream will be temporarily removed from the queue, if the `NGHTTP2_DATA_FLAG_EOF` flag is set, however, we signal that no more data will be available. If the `NGHTTP2_DATA_FLAG_NO_END_STREAM` flag is set, it means that we have trailing headers to send also.

addaleax: @jasnell Okay, thanks for the explanation! My question was actually a little more concrete – right now, `writable && stream_handle->queue_head_ == nullptr` cannot be true, I guess it should be `stream_handle->IsWritable() && stream_handle->queue_head_ == nullptr` if I’ve understood you correctly?

addaleax:
@jasnell Sorry, missed the thumbs up reaction. Anyway, my question there was silly, the code should work as-is. :smile:

jasnell:
:-) ok good. I've been traveling a bit the first part of this week so I haven't had a chance to dig into this in detail but I definitely will tomorrow (thursday). Really do appreciate what you've done here tho!

jasnell (approved):

jasnell:
@addaleax ... can you squash these down into a single commit? Then I'll get it landed. Thank you!

addaleax:
uh… kinda busy with something else right now :smile: sure I can do that, but feel free to just squash it while landing

jasnell:
:-) yeah, no worries, just if you have the time. I'm checking on the other PR that @mcollina opened at the moment. Once I'm done checking that I'll land this and squash if necessary :-)

jasnell:
Landed!

PR #43, opened 2017-02-22
http2: client round 1

mcollina:
@jasnell this is not working, I suspect it is not sending anything on the wire in fact.

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const debug = require('util').debuglog('http2');
 const assert = require('assert');
 const Buffer = require('buffer').Buffer;
 const EventEmitter = require('events');
@@ -35,6 +36,7 @@ const kSocket = Symbol('socket');
 const kState = Symbol('state');
 const kStreams = Symbol.for('streams');
 const kType = Symbol('type');
+const kProceed = Symbol('proceed');
 
 const kDefaultSocketTimeout = 2 * 60 * 1000;
 const kRenegTest = /TLS session renegotiation disabled for this socket/;
@@ -189,6 +191,11 @@ function request(headers, options) {
   unrefTimer(this);
   var handle = this[kHandle];
   options = options || {};
+  headers = headers || {};
+
+  if (headers[constants.HTTP2_HEADER_METHOD] === undefined)
+    headers[constants.HTTP2_HEADER_METHOD] = 'GET';
+
   if (typeof options !== 'object')
     throw new TypeError('options must be an object');
   if (options.weight === undefined)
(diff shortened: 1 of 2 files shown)


lib/internal/http2/core.js
```diff
@@ -757,6 +771,15 @@ class ServerHttp2Stream extends Http2Stream {
   }
 }
 
+ServerHttp2Stream.prototype._proceed = ServerHttp2Stream.prototype.respond;
```
jasnell: Let's use a Symbol instead of `_proceed`

jasnell (commented):
Quick read through... most bits look in order. Will pull it down and test in just a bit tho

mcollina:
@jasnell updated with `kProceed`.

jasnell:
+1... I started digging in yesterday then completely ran out of time to finish. Hoping to return to it this afternoon.

lib/internal/http2/core.js
```diff
@@ -571,7 +586,7 @@ class Http2Stream extends Duplex {
     unrefTimer(this);
     var state = this[kState];
     if (!state.headersSent)
-      this.respond();
+      this._proceed();
```
jasnell: `s/_proceed/[kProceed]` ??

mcollina: Ouch, yes. Done.

jasnell:
Ok, looks like there's a conflict with the change in node_http2.cc that needs to be looked at following @addaleax's refactor that landed yesterday. I'll look at resolving that in just a bit

mcollina:
@jasnell this is not working ATM. You can check out the branch as it is and let me know what the problem is.

This is my current client, and it's not working :(:

```js
'use strict'

const net = require('net');
const h2 = require('http2');

const socket = net.connect(8000, 'localhost');
const client = h2.createClientSession(socket);

const req = client.request();

socket.on('connect', function () {
  console.log('socket connected');
});

req.on('response', function (headers) {
  console.log('response!', headers)
})

let data = '';
req.setEncoding('utf8');
req.on('data', console.log);
```

jasnell:
Yeah, I've applied the patch locally and will be going through it in detail again here shortly. Hoping to have it working by the time we meet up in Seattle

jasnell:
Here's the "working" client 

```js
'use strict'

const common = require('../common');
const assert = require('assert');
const h2 = require('http2');
const net = require('net');
const body =
  '<html><head></head><body><h1>this is some data</h2></body></html>';

const server = h2.createServer();

server.on('stream', common.mustCall(onStream));

function onStream (stream) {
  stream.respond({
    'content-type': 'text/html',
    ':status': 200
  })
  stream.end(body);
}

server.listen(0);

server.on('listening', common.mustCall(function() {
  const socket = net.connect(this.address());
  const client = h2.createClientSession(socket);

  setImmediate(() => {
    const headers = {
      ':method': 'GET',
      ':scheme': 'http',
      ':authority': `localhost:${this.address().port}`,
      ':path': '/'
    };

    const req = client.request(headers);

    req.on('response', common.mustCall(function (headers) {
      console.log('response!', headers)
    }))

    let data = '';
    req.setEncoding('utf8');
    req.on('data', d => data += d);
    req.on('end', common.mustCall(function () {
      assert.strictEqual(body, data);
    }));
    req.end();
  });

}));
```

Also, in core.js, `createClientSession`, send the client settings in setImmediate: 

```js
  setImmediate(() => session.submitSettings(settings));
```

mcollina:
that's progress!
I've updated the test and it is _almost_ passing, but it's giving me a:

```
Assertion failed: (handle->ready_callbacks_head == nullptr), function nghttp2_session_free, file ../src/node-http2-core-inl.h, line 845.
Abort trap: 6
```

I'll work on it later on.

mcollina:
@jasnell this is ready to go. The test is green with a bunch of TODO!

jasnell:
Landed!

Issue #42, opened 2017-02-20
Test List


italoacasas:
cc: @jasnell, @mcollina, @addaleax

I start working on the test list today, I will be updating the list while gathering more information. I'm open the issue that way everyone can give feedback about what should be in/out/priority, etc...

## Test List

- [ ] Errors conditions (James list)
- [ ] Server with connections that prematurely exists (James list)
- [ ] Basic server with configuration options set (James list)
- [ ] Server that responds with a file (pipe test) (James list)
- [ ] Basic server with simple async response (James list)
- [ ] Basic server with and without TLS. Create server instance, start listening, accept request with simple sync response (James list)
- [ ] Server client error
- [ ] Response Status code
- [ ] HTTP Methods
- [ ] Invalid (URLs and headers fields)
- [ ] uncaughtException
- [ ] Default values (port, encoding, method, headers)
- [ ] Connect (request, response, length)
- [ ] DNS error

### Client

- [ ] abort event
- [ ] abort
- [ ] aborted event
- [ ] agent
- [ ] check HTTP token
- [ ] default headers exist
- [ ] default
- [ ] encoding
- [ ] get URL
- [ ] keep alive release before the finish
- [ ] parse error
- [ ] pipe end
- [ ] read in error
- [ ] readable
- [ ] reject chunked with content length
- [ ] response domain

mcollina:
hey @italoacasas, since today we have a minimal http2 client (see #43). The API is definitely not finalized, but we can definitely get some help in ramping up the test and test coverage. Maybe we can turn that bullet list into a task lists, and start adding those.
I think those tests would help in driving the client API.

cc @jasnell

italoacasas:
@mcollina I'm not going to be around next week.

-  I change the list for bullets point.

- The list is a work in progress, still pending most of the server tests. But at least we have a guide for a portion of what we need to do. I will work on this if no-one does when I come back.

- Maybe creating a Project is a good idea, remember the way we did this for the `streams internal states` tests?

mcollina:
@italoacasas definitely. We can do a project, and open some issues here.
It's a ton of work, as I expect those features *not* to work as expected (vs what we did for streams).

PR #47, opened 2017-03-23
Compatibility fixes and test cases

sebdeckers:
Accessing the request's headers was returning `undefined`. This is a quick fix that converts the headers object to a Map. Probably not the best for performance, but I wasn't sure if you'd rather rewrite all the property getters that expect `.has` and `.get` methods.

### DIFF
--- a/doc/http2-implementation-notes.md
+++ b/doc/http2-implementation-notes.md
@@ -50,18 +50,16 @@ const server = http2.createSecureServer(options, (req, res) => {
 
   res.writeHead(200, {'content-type': 'text/html'});
 
-  const favicon = res.createPushResponse();
-  favicon.path = '/favicon.ico';
-  favicon.push((req, res) => {
-    res.setHeader('content-type', 'image/jpeg');
-    fs.createReadStream('/some/image.jpg').pipe(res);
+  const favicon = res.createPushRequest({':path': '/favicon.ico'});
+  favicon.push((err, res) => {
+    res.setHeader('content-type', 'image/png');
+    fs.createReadStream('/some/logo.png').pipe(res);
   });
 
-  const pushResponse = res.createPushResponse();
-  pushResponse.path = '/image.jpg';
-  pushResponse.push((req, res) => {
+  const pushResponse = res.createPushRequest({':path': '/image.jpg'});
+  pushResponse.push((err, res) => {
     res.setHeader('content-type', 'image/jpeg');
-    fs.createReadStream('/some/image/jpg').pipe(res);
+    fs.createReadStream('/some/image.jpg').pipe(res);
   });
 
   res.end('<html><head><link rel="preload" href="/favicon.ico"/></head>' +
@@ -299,7 +297,7 @@ The `'rst-stream'` event is emitted when a RST-STREAM frame is received.
 ### Method: `response.writeHeader(statusCode, headers)`
 ### Method: `response.write()`
 ### Method: `response.end()`
-### Method: `response.createPushResponse()`
+### Method: `response.createPushRequest(headers)`
 
 ## HTTP2.createServerSession(options)
 
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -314,13 +314,17 @@ class Http2ServerResponse extends Stream {
   }
 
   addTrailers(headers) {
-    for (var name of Object.keys(headers)) {
-      name = String(name).trim().toLowerCase();
+    var trailers = this[kTrailers];
+    var keys = Object.keys(headers);
+    var key = '';
+    if (keys.length > 0)
+      return;
+    if (trailers === undefined)
+      trailers = this[kTrailers] = Object.create(null);
+    for (var i = 0; i < keys.length; i++) {
+      key = String(keys[i]).trim().toLowerCase();
       var value = headers[name];
       assertValidHeader(name, value);
-      var trailers = this[kTrailers];
-      if (trailers === undefined)
-        trailers = this[kTrailers] = Object.create(null);
       trailers[name] = String(value);
     }
   }
@@ -455,13 +459,19 @@ class Http2ServerResponse extends Stream {
     stream.respond(headers);
   }
 
-  createPushRequest(headers) {
+  createPushResponse(headers, callback) {
     var stream = this[kStream];
     if (stream === undefined) {
       throw new Error('HTTP/2 Stream has been closed');
     }
-    var request = new Http2PushRequest(stream, headers);
-    return request;
+    try {
+      stream.pushStream(headers, {}, function(stream, headers, options) {
+        var response = new Http2ServerResponse(stream);
+        callback(null, response);
+      });
+    } catch (error) {
+      callback(error);
+    }
   }
 
   [kBeginSend]() {
@@ -520,39 +530,6 @@ function onServerStream(stream, headers, flags) {
   }
 
   server.emit('request', request, response);
-}
-
-class Http2PushRequest {
-  constructor(stream, headers) {
-    this[kStream] = stream;
-    this[kHeaders] = headers || Object.create(null);
-    this[kState] = {
-      headersSent: false
-    };
-  }
-
-  push(callback) {
-    var state = this[kState];
-    if (state.headersSent === true)
-      throw new Error('Cannot push request headers ' +
-        'after the HTTP message has been sent');
-    var stream = this[kStream];
-    if (stream === undefined)
-      throw new Error('HTTP/2 Stream has been closed');
-    var headers = this[kHeaders];
-    try {
-      stream.pushStream(
-        headers,
-        function(promisedStream, headers, options) {
-          var response = new Http2ServerResponse(promisedStream);
-          callback(null, response);
-        }
-      );
-      state.headersSent = true;
-    } catch (error) {
-      callback(error);
-    }
-  }
 }
 
 module.exports = { onServerStream };
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -464,14 +464,10 @@ class Http2ServerResponse extends Stream {
     if (stream === undefined) {
       throw new Error('HTTP/2 Stream has been closed');
     }
-    try {
-      stream.pushStream(headers, {}, function(stream, headers, options) {
-        var response = new Http2ServerResponse(stream);
-        callback(null, response);
-      });
-    } catch (error) {
-      callback(error);
-    }
+    stream.pushStream(headers, {}, function(stream, headers, options) {
+      var response = new Http2ServerResponse(stream);
+      callback(null, response);
+    });
   }
 
   [kBeginSend]() {
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const linkedList = require('internal/linkedlist');
 const Stream = require('stream');
 const Readable = Stream.Readable;
 const binding = process.binding('http2');
@@ -22,31 +21,13 @@ const kTrailers = Symbol('trailers');
 // HTTP/2 implementation, intended to provide an interface that is as
 // close as possible to the current require('http') API
 
-function setHeader(list, name, value) {
-  name = String(name).toLowerCase().trim();
+function assertValidHeader(name, value) {
   if (isPseudoHeader(name))
     throw new Error('Cannot set HTTP/2 pseudo-headers');
   if (isIllegalConnectionSpecificHeader(name, value))
     throw new Error('Connection-specific HTTP/1 headers are not permitted');
   if (value === undefined || value === null)
     throw new TypeError('Value must not be undefined or null');
-  linkedList.append(list, [name, String(value)]);
-}
-
-function llistToHeaders(list, count) {
-  var ret = {};
-  while (!linkedList.isEmpty(list)) {
-    var item = list._idlePrev;
-    linkedList.remove(item);
-    var key = item[0];
-
-    if (ret[key]) {
-      ret[key].push(item[1]);
-    } else {
-      ret[key] = [item[1]];
-    }
-  }
-  return ret;
 }
 
 function isPseudoHeader(name) {
@@ -127,12 +108,14 @@ function onAborted(hadError, code) {
 }
 
 class Http2ServerRequest extends Readable {
-  constructor(stream, options) {
+  constructor(stream, headers, options) {
     super(options);
     this[kState] = {
+      statusCode: null,
       closed: false,
       closedCode: constants.NGHTTP2_NO_ERROR
     };
+    this[kHeaders] = headers;
     this[kStream] = stream;
     stream[kRequest] = this;
 
@@ -143,6 +126,9 @@ class Http2ServerRequest extends Readable {
     stream.on('error', onStreamError);
     stream.on('close', onStreamClosedRequest);
     stream.on('aborted', onAborted.bind(this));
+    const onfinish = this[kFinish].bind(this);
+    stream.on('streamClosed', onfinish);
+    stream.on('finish', onfinish);
     this.on('pause', onRequestPause);
     this.on('resume', onRequestResume);
     this.on('drain', onRequestDrain);
@@ -166,48 +152,33 @@ class Http2ServerRequest extends Readable {
     return this[kStream];
   }
 
+  get statusCode() {
+    return this[kState].statusCode;
+  }
+
   get headers() {
-    var stream = this[kStream];
-    return stream ? stream[kHeaders] : undefined;
+    return this[kHeaders];
   }
 
-  getHeader(name) {
-    var stream = this[kStream];
-    if (!stream)
-      return;
-    var headers = stream[kHeaders];
-    name = String(name).trim().toLowerCase();
-    return headers.get(name);
-  }
-
-  hasHeader(name) {
-    var stream = this[kStream];
-    if (!stream)
-      return false;
-    var headers = stream[kHeaders];
-    name = String(name).trim().toLowerCase();
-    return headers.has(name);
-  }
-
-  getTrailer(name) {
-    var trailers = this[kTrailers];
-    if (!trailers)
-      return;
-    name = String(name).trim().toLowerCase();
-    return trailers.get(name);
-  }
-
-  hasTrailer(name) {
-    var trailers = this[kTrailers];
-    if (!trailers)
-      return false;
-    name = String(name).trim().toLowerCase();
-    return trailers.has(name);
+  get rawHeaders() {
+    var headers = this[kHeaders];
+    if (headers === undefined)
+      return [];
+    var tuples = Object.entries(headers);
+    var flattened = Array.prototype.concat.apply([], tuples);
+    return flattened.map(String);
   }
 
   get trailers() {
-    var stream = this[kStream];
-    return stream ? stream[kTrailers] : undefined;
+    return this[kTrailers];
+  }
+
+  get httpVersionMajor() {
+    return 2;
+  }
+
+  get httpVersionMinor() {
+    return 0;
   }
 
   get httpVersion() {
@@ -224,43 +195,48 @@ class Http2ServerRequest extends Readable {
   }
 
   get method() {
-    var stream = this[kStream];
-    if (!stream) return;
-    var headers = stream[kHeaders];
-    return headers.get(constants.HTTP2_HEADER_METHOD);
+    var headers = this[kHeaders];
+    if (headers === undefined)
+      return;
+    return headers[constants.HTTP2_HEADER_METHOD];
   }
 
   get authority() {
-    var stream = this[kStream];
-    if (!stream) return;
-    var headers = stream[kHeaders];
-    return headers.get(constants.HTTP2_HEADER_AUTHORITY);
+    var headers = this[kHeaders];
+    if (headers === undefined)
+      return;
+    return headers[constants.HTTP2_HEADER_AUTHORITY];
   }
 
   get scheme() {
-    var stream = this[kStream];
-    if (!stream) return;
-    var headers = stream[kHeaders];
-    return headers.get(constants.HTTP2_HEADER_SCHEME);
+    var headers = this[kHeaders];
+    if (headers === undefined)
+      return;
+    return headers[constants.HTTP2_HEADER_SCHEME];
+  }
+
+  get url() {
+    return this.path;
   }
 
   get path() {
-    var stream = this[kStream];
-    if (!stream) return;
-    var headers = stream[kHeaders];
-    return headers.get(constants.HTTP2_HEADER_PATH);
+    var headers = this[kHeaders];
+    if (headers === undefined)
+      return;
+    return headers[constants.HTTP2_HEADER_PATH];
   }
 
   setTimeout(msecs, callback) {
     var stream = this[kStream];
-    if (!stream) return;
+    if (stream === undefined) return;
     stream.setTimeout(msecs, callback);
-    return this;
   }
 
   [kFinish](code) {
     var state = this[kState];
-    state.closeCode = code;
+    if (state.closed)
+      return;
+    state.closedCode = code;
     state.closed = true;
     this.push(null);
     this[kStream] = undefined;
@@ -273,6 +249,7 @@ class Http2ServerResponse extends Stream {
     this[kState] = {
       sendDate: true,
       statusCode: constants.HTTP_STATUS_OK,
+      headersSent: false,
       headerCount: 0,
       trailerCount: 0,
       closed: false,
@@ -280,23 +257,13 @@ class Http2ServerResponse extends Stream {
     };
     this[kStream] = stream;
     stream[kResponse] = this;
-    var headerList = {
-      _idleNext: null,
-      _idlePrev: null,
-    };
-    linkedList.init(headerList);
-    this[kHeaders] = headerList;
-    var trailersList = {
-      _idleNext: null,
-      _idlePrev: null,
-    };
-    linkedList.init(trailersList);
-    this[kHeaders] = trailersList;
     this.writable = true;
     stream.on('drain', onStreamResponseDrain);
     stream.on('error', onStreamResponseError);
     stream.on('close', onStreamClosedResponse);
     stream.on('aborted', onAborted.bind(this));
+    stream.on('streamClosed', this[kFinish].bind(this));
+    stream.on('finish', this[kFinish].bind(this));
   }
 
   get finished() {
@@ -336,7 +303,7 @@ class Http2ServerResponse extends Stream {
 
   set statusCode(code) {
     var state = this[kState];
-    if (state.headersSent)
+    if (state.headersSent === true)
       throw new Error('Cannot set status after the HTTP message has been sent');
     code |= 0;
     if (code >= 100 && code < 200)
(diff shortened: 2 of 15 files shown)


mcollina:
wooow, thanks! Can you add a quick unit test for this? You can see an example in: https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-create-client-session.js.

I'm not 100% sure what was the problem in the first place, so the test would help as well.

sebdeckers:
@mcollina The problem was that `stream[kHeaders]` as always `undefined`.

Actually I don't quite like the `Map` solution. I will rewrite this with `Object.prototype.hasOwnProperty` instead, so that the headers remain a plain object.

Will also work on a test case, sure. 😄

jasnell:
Why not pass the `headers` object off directly so that it does not need conversion?

```js
stream[kHeaders] = headers;
```

sebdeckers:
@jasnell Yep, doing that now. Got confused because the property getters use get/has. Not sure why...?

sebdeckers:
@jasnell Pushed a new commit that uses the plain object. No more conversion.

@mcollina Added a separate test script for compat. Made it extensible so I can also write test cases for other parts of that file (i.e. response).

sebdeckers:
PS: I see some unfinished provisions for implementing trailers on `Http2ServerRequest`. I believe this would involve listening to the `stream.trailers` event and storing them as `this[kTrailers]`. Possibly need to validate and merge multiple trailers. Is this a WIP? Happy to take a look at implementing this.

lib/internal/http2/core.js
```diff
@@ -441,6 +441,8 @@ function afterDoStreamWrite(status, handle, req) {
 
 function onHandleFinish() {
   var session = this[kSession];
+  if (!session)
```
sebdeckers: Btw this feels like a code smell. I found that `onHandleFinish` always runs twice. Second time the session is `undefined`, hence the guard.

sebdeckers:
@mcollina @jasnell I did some more work over the weekend to fix the compatibility layer. Please take a look and let me know what you think.

My aim is to make it work similar to Node.js' http APIs. Also added more tests, as requested. Probably covers some of #42.

- Am I going in the right direction with this PR?
- Anything I should clean up or refactor?
- How do you typically communicate on this project (Gitter, Slack, Hangouts)?

lib/internal/http2/core.js
```diff
@@ -1055,9 +1063,9 @@ function createClientSession(options, socket) {
   socket.on('pause', socketOnPause);
   socket.on('drain', socketOnDrain);
 
-  session.on('error', sessionOnError);
-  session.on('stream', sessionOnStream);
-  session.on('selectPadding', sessionOnSelectPadding);
+  // session.on('error', sessionOnError);
+  // session.on('stream', sessionOnStream);
+  // session.on('selectPadding', sessionOnSelectPadding);
```
mcollina: why those are commented?

lib/internal/http2/compat.js
```diff
@@ -330,45 +306,97 @@ class Http2ServerResponse extends Stream {
...
+        this.setHeader(key, headers[key]);
+      }
     }
     this.statusCode = statusCode;
     this[kBeginSend]();
-    return this;
+  }
+
+  writeHeader(statusCode, headers) {
+    return this.writeHead.apply(this, arguments);
```
mcollina: why this change? `apply(this, arguments)` is super slow.

lib/internal/http2/compat.js
```diff
@@ -330,45 +306,97 @@ class Http2ServerResponse extends Stream {
...
   }
 
+  addTrailers(headers) {
+    for (var name of Object.keys(headers)) {
+      name = String(name).trim().toLowerCase();
+      var value = headers[name];
+      assertValidHeader(name, value);
+      var trailers = this[kTrailers];
+      if (!trailers)
+        trailers = this[kTrailers] = Object.create(null);
```
mcollina: @jasnell can you check this? Is `Object.create(null)` ok?

lib/internal/http2/compat.js
```diff
@@ -428,20 +456,31 @@ class Http2ServerResponse extends Stream {
...
+    return stream.pushStream(headers, options, callback);
+  }
+
   [kBeginSend]() {
     var state = this[kState];
     var stream = this[kStream];
     if (!state.headersSent) {
       state.headersSent = true;
-      const headers = llistToHeaders(this[kHeaders]);
+      const headers = this[kHeaders] || Object.create(null);
```
mcollina: @jasnell I think `llistToHeaders` is needed, or am I missing something?

mcollina (changes requested):
Some questions around the code, also can you split the test into multiple files? I would prefer it as they are much easier to debug.

lib/internal/http2/compat.js
sebdeckers: TIL `apply`/`call` are about [half the speed](https://jsperf.com/function-calls-direct-vs-apply-vs-call-vs-bind/6) of a direct call on Chrome.

Saw writeHeaders in `_http_server.js` being forwarded to writeHead [exactly this way using `apply`](https://github.com/nodejs/http2/blob/master/lib/_http_server.js#L232-L234).

Looks like [the writeHeaders API has been deprecated](https://github.com/nodejs/node/commit/fb71ba49212dd21c0404892b4bddc37437bd5397) and dropped from the main nodejs/node repo.

Will remove it.

lib/internal/http2/core.js
sebdeckers: These are event handlers for the server. They each use `session[kServer]` which is undefined for clients. I was wondering if we need these handlers at all, or if they should do something else perhaps. Thoughts?

jasnell: These are not just for the server side. Each of these will be relevant for the client side also.

jasnell:
I'll take a more detailed look through in the morning.

lib/internal/http2/compat.js
sebdeckers: Looking at a [jsPerf test case](https://jsperf.com/object-create-null-vs-literal/2), it seems `{}` is 10-20x faster than `Objet.create(null)`. Should change it to that, I guess.

lib/internal/http2/core.js
sebdeckers: @jasnell These events are certainly relevant to the client as a public interface. I used the `session:stream` event in one of the tests. But these internally assigned handlers were server-specific. It is not clear to me how this should be rewritten. Suggestion?

lib/internal/http2/compat.js
jasnell: `Object.create(null)` is better in this case.

sebdeckers:
@mcollina Thank you for the feedback! 🙇 Have now separated the tests into separate files. Also linted the code.

Notes to self:

Run Eslint as per the Makefile:
```
./node tools/eslint/bin/eslint.js --cache --rulesdir=tools/eslint-rules "test/parallel/test-http2-*"
```

Quickly run the relevant test cases in bash:
```
for test in test/parallel/test-http2-*.js; do ./node $test; done
```

sebdeckers:
WIP:

- Cookies are not currently supported the same way the http1 module handles them.

- createPushResponse does not yet follow the API design shown in the [implementation nodes](https://github.com/nodejs/http2/blob/master/doc/http2-implementation-notes.md#example). For now it simply returns the ServerHttp2Stream.

mcollina:
> Cookies are not currently supported the same way the http1 module handles them.

can you please open a separate issue about cookies?

sebdeckers:
In 373969563 I've implemented a minimalist API for push requests and push responses.

```js
server.on('request', (request, response) => {
  const headers = {
    ':path': '/foobar.txt',
    'content-type': 'text/plain; charset=utf8'
  }

  const pushRequest = response.createPushRequest(headers)
  // createPushRequest returns an Http2PushRequest
  // Its argument is a plain Object with the PUSH_PROMISE headers (key/value).

  // Its only method is .push() which sends the PUSH_PROMISE frame,
  // and offers a callback to deal with the push response.
  pushRequest.push((error, pushResponse) => {
    // pushResponse is just an Http2ServerResponse,
    // so it can do all the usual tricks...
    pushResponse.writeHead(200, headers)
    pushResponse.end('Hello, World!')
  })
})
```

@jasnell Would you be okay with this design replacing the one from the [implementation notes example](https://github.com/nodejs/http2/blob/76dc88b4d9a140497a58559d0882c76cc5616913/doc/http2-implementation-notes.md#example)? It omits the `.path` property and `request` callback argument. Also renamed it to `createPushRequest` since that aligns with the RFC terminology (push request -> push response).

~~@mcollina I'm struggling to test the push stream using the client API. It looks like the `PUSH_PROMISE` frame is received in `onSessionHeaders` as a `NGHTTP2_HCAT_RESPONSE`, with the stream ID of the parent stream, i.e. the ID of the stream that received the frame, instead of the promised stream ID. Either I'm missing something or this may need some work on the C++ side. I'm not so good at that yet so any help is welcome. (I manually tested the server side's `createPushRequest` code using nghttp client. Pretty sure the bug is in the client side code.)~~ Update: Fixed, see https://github.com/nodejs/http2/pull/47#issuecomment-290984833

sebdeckers:
@mcollina @jasnell Please review the C++ changes in https://github.com/nodejs/http2/pull/47/commits/7f8a5ebe441f66e1d7a84f05ea07404948bf3762#diff-20934af26ea78010b56b52f2fc636a18

This commit adds support for promised stream ID when receiving headers. The test case in `test-http2-compat-Http2ServerResponse-createPushRequest.js` passes, so push streams can now be used in the server and client.

However I'm seeing this error on exit:

```
$ ./node test/parallel/test-http2-compat-Http2ServerResponse-createPushRequest.js
Assertion failed: (pending_callbacks_head_ == nullptr), function Free, file ../src/node-http2-core-inl.h, line 325.
Abort trap: 6
```

Any ideas? It's been 20 years since I touched C++. This has been another adventurous weekend. 😅

*Update:* Added a workaround inn 44cc0b8. By delaying the server's `pushStream.end()`, the client reads the data before closing its stream. I think this bug should be addressed in the context of #51 with a minimal test case, rather than this PR which is focused on server changes.

lib/internal/http2/compat.js
sebdeckers: TIL: The [nghttp2 `nghttp2_on_header_callback` API](https://nghttp2.org/documentation/types.html#c.nghttp2_on_header_callback) fires one header (name/value pair) at a time, so it is reasonable to append decoded headers to a (linked) list as they are received. This may explain why the linked list was used on the JS side.

sebdeckers:
@mcollina @jasnell Any chance to get this PR reviewed & merged/rejected? Happy to do a Hangout call and discuss ways to collaborate. I'm seeing the API changes in #51 and am happy to update this PR accordingly, either in this PR or after round 11 closes. Thanks!

jasnell:
This PR will need to be reworked. I've updated the master against the latest core and squashed the numerous http2 related commits to clean things up and get us ready for the next phase of cleanup and adding tests, etc.

sebdeckers:
@jasnell Great! I'll get to work on it. 🕺

sebdeckers:
@jasnell I have reworked the code according to the new client API. Also rebased/squashed and then cherry-picked the patch as one commit onto the updated master branch. Hope this is acceptable.

😫 Apologies for losing the individual commits — I struggle with advanced git-fu.

doc/http2-implementation-notes.md
```diff
@@ -50,21 +50,19 @@ const server = http2.createSecureServer(options, (req, res) => {
```
jasnell, line 50: Likely best to leave out the changes to this file... these implementation notes are out of date and will be replaced entirely very soon by proper documentation.

lib/internal/http2/compat.js
```diff
@@ -223,43 +194,48 @@ class Http2ServerRequest extends Readable {
   }
 
   get method() {
-    var stream = this[kStream];
-    if (!stream) return;
-    var headers = stream[kHeaders];
-    return headers.get(constants.HTTP2_HEADER_METHOD);
+    var headers = this[kHeaders];
+    if (!headers)
```
jasnell: I believe core is generally moving towards explicitly checking against `=== undefined` with these types of if statement

lib/internal/http2/compat.js
```diff
@@ -279,13 +255,13 @@ class Http2ServerResponse extends Stream {
...
     stream[kResponse] = this;
-    this[kHeaders] = linkedList.create();
-    this[kTrailers] = linkedList.create();
     this.writable = true;
     stream.on('drain', onStreamResponseDrain);
     stream.on('error', onStreamResponseError);
     stream.on('close', onStreamClosedResponse);
     stream.on('aborted', onAborted.bind(this));
+    stream.on('streamClosed', this[kFinish].bind(this));
+    stream.on('finish', this[kFinish].bind(this));
```
jasnell, line 266: Likely better to bind the function once then attach it to both...

```js
const onfinish = this[kFinish].bind(this);
stream.on('streamClosed', onfinish);
stream.on('finish', onfinish);
```

lib/internal/http2/compat.js
```diff
@@ -330,45 +306,94 @@ class Http2ServerResponse extends Stream {
     code |= 0;
     if (code >= 100 && code < 200)
       throw new RangeError('Informational status codes cannot be used');
-    if (code < 200 || code > 999)
+    if (code < 200 || code >= 600)
```
jasnell: The `999` here is intentional for compatibility with the existing core implementation.

https://github.com/nodejs/node/blob/master/lib/_http_server.js#L187

lib/internal/http2/compat.js
```diff
@@ -330,45 +306,94 @@ class Http2ServerResponse extends Stream {
...
       throw new RangeError(`Invalid status code: ${code}`);
     state.statusCode = code;
   }
 
+  addTrailers(headers) {
+    for (var name of Object.keys(headers)) {
```
jasnell: I need to see if there are other areas where this needs to be changed, but please use a tradition `for (var n = 0; ...` style here. They out perform the `for-of` and `for-in` loops significantly.

lib/internal/http2/compat.js
jasnell: The more I think about this, the more I'm thinking that we may want to go ahead and make the jump to use `Map`. The performance profile is significantly improved and should be better by the time this lands.

lib/internal/http2/compat.js
jasnell: yes, this is exactly why a linked list was used. I'll have to dig in a bit more to see what is best here.

lib/internal/http2/compat.js
```diff
@@ -486,4 +520,41 @@ function onServerStream(stream, headers, flags) {
   server.emit('request', request, response);
 }
 
+class Http2PushRequest {
```
jasnell: Will have to give this some more thought. At this point, at least for the early versions of this, I'm not convinced that we should expose push support via the compat API. Still stewing tho

lib/internal/http2/core.js
jasnell: Yep, this was an area that I've been planning to get back to (specifically, the sequencing of the events and a well defined event lifecycle)

lib/internal/http2/core.js
jasnell: will be thinking through this bit more now that time is freeing up to work on it again :-) hopefully more suggestions soon.

test/parallel/test-http2-binding.js
```diff
@@ -13,7 +13,7 @@ assert(http2.SessionShutdownWrap);
 assert.strictEqual(typeof http2.Http2Session, 'function');
 assert.strictEqual(typeof http2.SessionShutdownWrap, 'function');
 
-const settings = http2.getDefaultSettings({});
+const settings = require('http2').getDefaultSettings({});
```
jasnell: Why this change?

test/parallel/test-http2-compat-Http2ServerRequest-headers.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
```
jasnell: btw, test filenames should always be all lower case.

test/parallel/test-http2-compat-Http2ServerRequest.js
```diff
@@ -0,0 +1,58 @@
...
+  const url = `http://localhost:${server.address().port}`;
+  const client = h2.connect(url, common.mustCall(function() {
+    const headers = {
+      ':path': '/foobar',
+      ':method': 'GET',
+      ':scheme': 'http',
+      ':authority': 'localhost'
+    };
+    const request = client.request(headers);
+    // request.on('response', common.mustCall(function(headers) {
```
jasnell: Why are these commented out? If they are todo for later, please include a `TODO(sebdeckers):` comment with a quick explanation

jasnell (changes requested):
Ok, first pass done. I'm going to be applying the patch and kicking the tires on it on the flight over to berlin. Hopefully will have more comments after that! Sorry it's taken so long!

lib/internal/http2/compat.js
sebdeckers: Does V8 5.8, [as tested with Chrome v58](https://jsperf.co/for-vs-foreach-vs-for-of), not eliminate the performance difference?

chrome for/for-of benchmark

I can still see a 30% advantage in microbenchmarks on Node.js (this fork) with V8 5.7 using `for` loop over `for-of`. Gist: https://gist.github.com/sebdeckers/f110f4c5f65fc70da39587cb5d0fba68

```
for x 29,955,989 ops/sec ±1.58% (81 runs sampled)
for-of x 22,832,152 ops/sec ±1.64% (78 runs sampled)
```

#### Update

Using Node.js canary (v8: 6.0.125) there's a performance regression for both loop styles. Weird.

```
for x 26,097,145 ops/sec ±1.67% (82 runs sampled)
for-of x 17,436,809 ops/sec ±1.32% (77 runs sampled)
```

lib/internal/http2/compat.js
sebdeckers: FWIW my thoughts about the push API and compat layer.

1. Both npm:http2 and npm:spdy expose a compatibility layer with added support for PUSH_PROMISE.

   Quote: https://www.npmjs.com/package/http2#api
   > The goal is the perfect API compatibility, with additional HTTP2 related extensions (like server push).

1. I renamed it to `createPushRequest` to match ["Push Request" in the RFC](http://httpwg.org/specs/rfc7540.html#PushRequests), meaning the PUSH_PROMISE frame. The HEADERS/DATA frames are collectively referred to as the "Push Response," whose API is exactly the same as a regular `Http2ServerResponse`.

1. As a user of the H2 APIs it's important for me to be able to send the push requests (PUSH_PROMISE) before the main response (HEADERS/DATA for the user's request), followed finally by the push responses (HEADERS/DATA for the PUSH_PROMISE). This is the optimal flow described in: https://en.wikipedia.org/wiki/HTTP/2_Server_Push

test/parallel/test-http2-binding.js
sebdeckers: This test was failing otherwise. Now matches the code just below: https://github.com/sebdeckers/http2/blob/430f9a648163b612a2240151b41e3639c4a44958/test/parallel/test-http2-binding.js#L28

test/parallel/test-http2-compat-Http2ServerRequest-headers.js
sebdeckers: Ah right :) I knew something looked out of place... will rename them.

test/parallel/test-http2-compat-Http2ServerRequest.js
sebdeckers: Mea culpa. Forgot to clean up these development artefacts.

sebdeckers:
- Updated with the latest changes in `master`.
- Renamed test files to lowercase.
- Acceptable status code limit increased to 999 & added a test case for this.
- Explicitly check for `=== undefined` or `=== false` instead of using `if (!...)` shorthand.

jasnell:
Tied up in the collaborator summit today but will go back through this in more detail very soon

robertkowalski:
ah wow, this is why i can't access headers today when trying to port tests from http to http2!

jasnell:
yeah, once we get this landed headers should work as needed. I should be able to get the full review done tomorrow :-)

lib/internal/http2/core.js
```diff
@@ -848,7 +863,7 @@ class ServerHttp2Stream extends Http2Stream {
       options.readable = !options.endStream;
       const stream = new ServerHttp2Stream(session, ret, options);
       streams.set(ret, stream);
-      process.nextTick(callback, stream, headers, 0);
+      process.nextTick(callback, null, stream, headers, 0);
```
jasnell: I assume the `null` here is to make it match the typical Node.js callback `(err, ...args)` pattern? If so, I do not believe that's necessary. The objects involved here are event emitters that will emit `error` events when errors occur.

lib/internal/http2/core.js
```diff
@@ -1202,8 +1219,8 @@ function createClientSession(options, socket) {
   socket.on('drain', socketOnDrain);
 
   session.on('error', clientSessionOnError);
-  session.on('stream', sessionOnStream);
-  session.on('selectPadding', sessionOnSelectPadding);
+  // session.on('stream', sessionOnStream);
+  // session.on('selectPadding', sessionOnSelectPadding);
```
jasnell: For handling these, client specific handlers can be used, similar to how a few other client specific events are handled. I'd have to take a look more in depth to see what the changes would need to be. Would you mind adding a `TODO` comment on these to provide client specific implementations?

test/parallel/test-http2-binding.js
jasnell: this part was fixed in a separate PR. After a rebase you should be able to drop this change.

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,96 @@
...
+function isOdd(number) {
+  return number % 2 === 1;
+}
+
+// Push a request & response
+
+const server = h2.createServer();
+server.listen(0, common.mustCall(function() {
+  server.once('request', common.mustCall(function(request, response) {
+    assert.ok(isOdd(response.stream.id));
```
jasnell: Since this is only used once, perhaps best to just inline the `response.stream.id % 2 === 1` check

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,96 @@
...
+      ':method': 'GET',
+      ':scheme': 'http',
+      ':authority': 'localhost'
+    };
+
+    const pushRequest = response.createPushRequest(headers);
+
+    pushRequest.push(common.mustCall(function(error, pushResponse) {
+      assert.strictEqual(error, null);
+      assert.ok(isEven(pushResponse.stream.id));
```
jasnell: ditto here... the separate `isEven()` and `isOdd()` methods can just be inlined.

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,96 @@
...
+
+      pushResponse.write('This is a server-initiated response');
+
+      // TODO(sebdeckers) Remove this forced delay workaround.
+      // pushResponse.end();
+      // response.end();
+      setTimeout(() => pushResponse.end(), 100);
+      setTimeout(() => response.end(), 200);
```
jasnell: Quick reminder for me... why is this delay needed again? I may have missed an earlier explanation

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,96 @@
...
+    }));
+  }));
+
+  const url = `http://localhost:${server.address().port}`;
+  const client = h2.connect(url, common.mustCall(function() {
+    const headers = {
+      ':path': '/',
+      ':method': 'GET',
+      ':scheme': 'http',
+      ':authority': 'localhost'
```
jasnell: `:authority` should include the port also... so `':authority': `localhost:${server.address().port}` here

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,96 @@
...
+    };
+
+    const requestStream = client.request(headers);
+
+    function onStream(pushStream, headers, flags) {
+      assert.strictEqual(headers[':path'], '/pushed');
+      assert.strictEqual(headers[':method'], 'GET');
+      assert.strictEqual(headers[':scheme'], 'http');
+      assert.strictEqual(headers[':authority'], 'localhost');
```
jasnell: just as an aside, these standard header names are exported as constants

test/parallel/test-http2-compat-serverrequest.js
```diff
@@ -0,0 +1,45 @@
...
+  const client = h2.connect(url, common.mustCall(function() {
+    const headers = {
+      ':path': '/foobar',
+      ':method': 'GET',
+      ':scheme': 'http',
+      ':authority': 'localhost'
+    };
+    const request = client.request(headers);
+    request.on('end', common.mustCall(function() {
+      client.socket.destroy();
```
jasnell: Note: Once https://github.com/nodejs/http2/pull/64 lands, this can be simplified to just `client.destroy()`

jasnell (changes requested):
Getting closer! Left a few comments and questions. Great work on this. We should be able to get it landed soon.

test/parallel/test-http2-compat-pushrequest.js
sebdeckers: Ah yes, I was wondering why the constants are preferred. What's the difference?

test/parallel/test-http2-compat-pushrequest.js
sebdeckers: Good catch! This bug has eluded me. Seemed like the stream was being destroyed before all data had been sent/read. Possibly related to #72 and/or #77.

sebdeckers:
- Rebased on master
- Added `TODO(jasnell)` note for client event handlers (core)
- Inlined `isOdd`/`isEven` (test)
- Include port in request `:authority` header (tests)
- Made `state.headersSent === true|false` checks explicit, caught a bug (compat)
- Fixed inverted headersSent check in `Http2PushRequest.push()` (compat)
- Simplified cleanup to `client.destroy()` (tests)

test/parallel/test-http2-compat-pushrequest.js
```diff
@@ -0,0 +1,88 @@
...
+      ':authority': `localhost:${server.address().port}`
+    };
+
+    const requestStream = client.request(headers);
+
+    function onStream(pushStream, headers, flags) {
+      assert.strictEqual(headers[':path'], '/pushed');
+      assert.strictEqual(headers[':method'], 'GET');
+      assert.strictEqual(headers[':scheme'], 'http');
+      assert.strictEqual(headers[':authority'], `localhost:${server.address().port}`);
```
jasnell: nit: long line here.

jasnell (approved):
LGTM. If you run make lint there will be some errors but they are from other things and I'm fixing them in a separate PR. There is one linting issue in http2-compat-pushrequest. I'd like @mcollina to sign off on this too since he's leading the compat.js side.

sebdeckers:
Thanks for the thorough reviews, @jasnell. ☺️ I've cleaned up the linting issue and rebased once more.

lib/internal/http2/compat.js
mcollina: can you please do this with a `for` loop?

lib/internal/http2/compat.js
```diff
@@ -497,4 +522,37 @@ function onServerStream(stream, headers, flags) {
...
+          var response = new Http2ServerResponse(promisedStream);
+          callback(null, response);
+        }
+      );
+      state.headersSent = true;
+    } catch (error) {
+      callback(error);
+    }
+  }
+}
```
mcollina: Would it be better to put all the logic in `createPushRequest()`?  Why do we need this `class`?

I promose we move the API from:

```js
const headers = {
   ':path': '/pushed',
   ':method': 'GET',
   ':scheme': 'http',
   ':authority': `localhost:${port}`
 };
 
const pushRequest = response.createPushRequest(headers);

pushRequest.push(common.mustCall(function(error, pushResponse) {
}))
```

to:

```js
const headers = {
   ':path': '/pushed',
   ':method': 'GET',
   ':scheme': 'http',
   ':authority': `localhost:${port}`
 };
 
response.createPushRequest(headers, common.mustCall(function(error, pushResponse) {
}))
```

test/parallel/test-http2-compat-pushrequest.js
mcollina: can you please add a TODO with a link the above bugs?

mcollina (changes requested):
This is ok in the implementation, with a few nits. I do not agree with the proposed API, and I think it could be highly simplified.

lib/internal/http2/compat.js
sebdeckers: Agreed. I don't know what I was thinking. 🤦‍♂️

c6a2fe2 implements the API as requested. Also renamed to `createPushResponse` since the returned object is an `Http2ServerResponse`.

lib/internal/http2/compat.js
```diff
@@ -439,20 +459,37 @@ class Http2ServerResponse extends Stream {
...
+      throw new Error('HTTP/2 Stream has been closed');
+    }
+    try {
+      stream.pushStream(headers, {}, function(stream, headers, options) {
+        var response = new Http2ServerResponse(stream);
+        callback(null, response);
+      });
+    } catch (error) {
+      callback(error);
+    }
```
mcollina: why are we catching this?

sebdeckers: Habit from projects with other coding styles. Changed it up to simply throw instead.

mcollina:
@sebdeckers sorry this conflicted with @yosuke-furukawa change. Would any of you fix things?

sebdeckers:
@yosuke-furukawa Does the change in 6765d00 work for you? (See also [`onSessionHeaders`]( https://github.com/nodejs/http2/pull/47/commits/ed77c36ef6ac9679677c8e41de8a7b9b8ebea679#diff-696b2cc418addca5f3fe5020058f8b15R93))

The `stream` event is emitted by both server and client for any incoming streams. E.g. server receives a request, or client receives a push request. This is where request headers can be handled.

The `push` event is now emitted on the client's push request stream. This allows matching of the push response to the corresponding push request.

mcollina (approved):

mcollina:
Landed as cee5479

jasnell:
@sebdeckers ... thank you for bearing with us on this. I know it took a while to get landed.

sebdeckers:
@jasnell I learned a lot by doing this and look forward to making more contributions. My heartfelt thanks to you and @mcollina for all the patience & guidance. 🥂

PR #48, opened 2017-03-23
response.createPushResponse

### LINKED ISSUE #46: Cannot find res.createPushResponse() in Http2Response.
* **v8.0.0-pre**:
* **Darwin**:

Thanks for your nice work.

Seems that the [doc](https://github.com/nodejs/http2/blob/master/doc/http2-implementation-notes.md#method-responsecreatepushresponse) for the `http2` is not updated, I know that the `http2` module itself implemented as a separate module named `http2` but cannot find the `res.createPushResponse` method. Can you please help me?

Thanks again.

sebdeckers:
As per the [HTTP/2 API documentation](https://github.com/nodejs/http2/blob/master/doc/http2-implementation-notes.md#method-responsecreatepushresponse), the `createPushResponse` method should be exposed on the response object.

It takes the same arguments as `Stream.pushStream`. Not sure if that is the intention?

Fixes #46

### DIFF
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -428,6 +428,15 @@ class Http2ServerResponse extends Stream {
     stream.respond(headers);
   }
 
+  createPushResponse(headers, options, callback) {
+    var stream = this[kStream];
+    if (!stream) {
+      callback(new Error('HTTP/2 Stream has been closed'));
+      return;
+    }
+    return stream.pushStream(headers, options, callback);
+  }
+
   [kBeginSend]() {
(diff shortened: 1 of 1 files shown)


mcollina:
Can you add a quick unit test for this? You can see an example in: https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-create-client-session.js.

sebdeckers:
Closing this PR in favour of #47. I've added an end to end test case in 12e856b. Turned out a few other things in the event flow also needed fixing to really make push streams work correctly.

PR #49, opened 2017-03-23
Adding request.url for connect compatibility

sebdeckers:
Having the `request.url` property aliased to `request.path` fixes a bug with the popular connect framework. Is this acceptable as backwards compatibility or would it be API bloat?

### DIFF
--- a/test/parallel/test-http2-create-client-session.js
+++ b/test/parallel/test-http2-create-client-session.js
@@ -6,8 +6,15 @@ const h2 = require('http2');
 const net = require('net');
 const body =
   '<html><head></head><body><h1>this is some data</h2></body></html>';
+let pathname;
 
-const server = h2.createServer();
+const server = h2.createServer(common.mustCall(requestListener));
+
+function requestListener(request, response) {
+  assert.equal(pathname, request.path);
+  assert.equal(pathname, request.url);
+  assert.equal(pathname, request.headers[':path']);
+}
 
 // we use the lower-level API here
 server.on('stream', common.mustCall(onStream));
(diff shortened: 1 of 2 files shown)


mcollina:
Definitely, but would you mind adding a unit test?See https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-create-client-session.js as an example of a client API.

sebdeckers:
@mcollina I'd be happy to. Still figuring out how to efficiently develop on the Node codebase. Takes a while to recompile.
- Is there like a "watch mode" that monitors source code changes, recompiles, and re-runs tests more efficiently?
- Do I really need to recompile (`make && make install`) even if I only change the JS files?

mcollina:
no there is not a "watch mode", but we usually just run a single unit test, so: `make && ./node https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-create-client-session.js ` in the example. If it does not throw, it passes.

You do not have to do `make install`.

It does not recompile all of it if you do not switch between branches.

sebdeckers:
@mcollina @jasnell Would the test case in the second commit (967a0fb) be suitable? I just mean the way it's written; the test won't actually pass until to the request headers problem is resolved in #47.

test/parallel/test-http2-create-client-session.js
```diff
@@ -28,12 +35,14 @@ server.on('listening', common.mustCall(function() {
...
     const client = h2.createClientSession(socket);
+    
+    pathname = `/${Math.random()}`;
 
     const headers = {
       ':method': 'GET',
       ':scheme': 'http',
       ':authority': `localhost:${this.address().port}`,
-      ':path': '/'
+      ':path': pathname
```
mcollina, line 45: can you please create a new test for this?

sebdeckers: @mcollina I could add a test case for this fix to the test script in #47. So would need that to land in master branch first, then update this PR with the test case.

sebdeckers:
Closing this PR in favour of #47. I've added the same `url` -> `path` alias, with a test assertion, in 326fee5.

PR #51, opened 2017-03-29
Round 11

jasnell:
beginning of round 11, starting with some API refinement

### DIFF
--- a/lib/http2.js
+++ b/lib/http2.js
@@ -11,5 +11,6 @@ module.exports = {
   createServer: core.createServer,
   createSecureServer: core.createSecureServer,
   createServerSession: core.createServerSession,
-  createClientSession: core.createClientSession
+  connect: core.connect,
+  secureConnect: core.secureConnect
 };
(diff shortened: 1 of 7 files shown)


mcollina (commented):
I'm actually 👎  on this change, if it is possible.

In the rest of core we have:

```js
const sock = net.connect()
```

while here we will have:

```js
h2.connect(url, function (err, client) {

})
```

I think we should strive to implement the delayed open pattern in h2, it will simplify usage greatly. However, if it is unfeasible, we would have to implement an agent-like behavior on top.

jasnell:
Understood. This is incremental while I work out the requirements on a few other bits. The delayed open pattern is going to be a bit of a larger internal change so I'm pushing it off slightly

mcollina:
Good job! I like having just a single `connect()`.

jasnell:
I should have another round of updates on this later today

mcollina:
@jasnell that execution bit looks really bad on the files.

jasnell:
Ugh. I keep forgetting about that. Will switch it up :)

jasnell:
Going to get this landed in master, then open a PR that updates with the current upstream master

jasnell:
Landed

PR #53, opened 2017-04-03
Fix wrong markdown syntax of badge url

blade254353074:
Just a markdown syntax fix contribution 😂

##### Checklist

- [ ] commit message follows commit guidelines

Nothing


mcollina:
Thanks, but this is a branch cut from nodejs/node. We are rebasing it periodically on top of that. So, there is no need for any readme fix

blade254353074:
Okay

Issue #55, opened 2017-04-26
Client hangs after first WINDOW_UPDATE


kjin:
The client seems to hang sometimes after sending its first `WINDOW_UPDATE` frame, and this is most readily seen when it opens a single stream with a small `initialWindowSize`. In C++, `DrainCallbacks` isn't being called after the first window update frame is sent.

I think there is a race condition involved, given that opening multiple streams increases the likelihood that all streams will receive all of their data without hanging, probably because the interleaving gives each stream enough time to reset state or something.

Maybe this is related to #19?

jasnell:
@kjin ... is this resolved now?

jasnell:
I believe this is resolved now. @kjin, if not, I'll reopen

PR #56, opened 2017-04-28
Enable auto window size updates

kjin:
This is an attempt to address #55, as manually specifying the amount of data consumed seem to cause a receiver not to emit WINDOW_UPDATE frames in some cases.

It seemed like disabling auto window updates was an intentional decision, though, so I'd be interested in knowing why this was the case.


jasnell:
It was intentionally done to give us a bit of flexibility while prototyping, benchmarking and determining if a manual strategy would be more efficient. Given the experience so far, however, I think allowing nghttp to handle it automatically is the better strategy.

jasnell (approved):

mcollina (approved):
LGTM

kjin:
@jasnell Ok, thanks for the explanation!

jasnell:
Landed

PR #60, opened 2017-05-02
Change stream write loop exit condition in Nghttp2Session::OnStreamRead

kjin:
The loop that writes queued write requests into a buffer in `Nghttp2Session::OnStreamRead` should break when either there is no more data to write, or the buffer fills up. It seems like it didn't do the latter originally. This becomes apparent when multiple buffers of unwritten bytes are queued in a single write request that don't all fit in `buf`: the for loop in lines 190-209 will always exhaustively iterate through all `nbufs` buffers without writing them, and then the outer while loop will move onto the next element in `queue_head_`.

This change makes the loop exit when no more bytes can be written.

I have tested this manually and will upload a test shortly.

### DIFF
--- a/test/parallel/test-http2-client-window-size.js
+++ b/test/parallel/test-http2-client-window-size.js
@@ -1,0 +1,100 @@
+'use strict';
+
+// This test ensures that servers are able to send data independent of window
+// size.
+
+const assert = require('assert');
+const common = require('../common');
+const h2 = require('http2');
+
+// Given a list of buffers and an initial window size, have a server write
+// each buffer to the HTTP2 Writable stream, and let the client verify that
+// all of the bytes were sent correctly
+function run(buffers, initialWindowSize) {
+  return new Promise((resolve, reject) => {
+    const expectedBuffer = Buffer.concat(buffers);
+
+    const server = h2.createServer();
+    const a = [];
+    server.on('stream', (stream) => {
+      let i = 0;
+      const writeToStream = () => {
+        let cont = () => {
+          i++;
+          if (i < buffers.length) {
+            setImmediate(writeToStream);
+          } else {
+            stream.end();
+          }
+        }
+        const drained = stream.write(buffers[i]);
+        if (drained) {
+          cont();
+        } else {
+          stream.once('drain', writeToStream);
+        }
+      }
+      writeToStream();
+    });
+    server.listen(0);
+
+    server.on('listening', common.mustCall(function() {
+      const port = this.address().port;
+
+      const client =
+        h2.connect({
+          authority: 'localhost',
+          protocol: 'http:',
+          port
+        }, {
+          settings: {
+            initialWindowSize
+          }
+        }).on('connect', common.mustCall(() => {
+          const req = client.request({
+            ':method': 'GET',
+            ':path': '/'
+          });
+          let responses = [];
+          req.on('data', data => {
+            responses.push(data);
+          });
+          req.on('end', common.mustCall(() => {
+            const actualBuffer = Buffer.concat(responses);
+            assert.strictEqual(Buffer.compare(actualBuffer, expectedBuffer), 0);
+            // shut down
+            client.socket.destroy();
+            server.close(() => {
+              resolve();
+            });
+          }));
+          req.end();
+        }));
+    }));
+  });
+}
+
+const bufferValueRange = [0, 1, 2, 3];
+const buffersList = [
+  bufferValueRange.map(a => Buffer.alloc(1 << 4, a)),
+  bufferValueRange.map(a => Buffer.alloc((1 << 8) - 1, a)),
+  bufferValueRange.map(a => Buffer.alloc(1 << 8, a)),
+  bufferValueRange.map(a => Buffer.alloc(1 << 17, a))
+];
+const initialWindowSizeList = [
+  1 << 4,
+  (1 << 8) - 1,
+  1 << 8,
+  1 << 17,
+  undefined // use default window size which is (1 << 16) - 1
+];
+
+// Call `run` on each element in the cartesian product of buffersList and
+// initialWindowSizeList.
+let p = Promise.resolve();
+for (const buffers of buffersList) {
+  for (const initialWindowSize of initialWindowSizeList) {
+    p = p.then(() => run(buffers, initialWindowSize));
+  }
+}
+p.then(common.mustCall(() => {}));
--- a/test/parallel/test-http2-window-size.js
+++ b/test/parallel/test-http2-window-size.js
@@ -15,25 +15,24 @@ function run(buffers, initialWindowSize) {
     const expectedBuffer = Buffer.concat(buffers);
 
     const server = h2.createServer();
-    const a = [];
     server.on('stream', (stream) => {
       let i = 0;
       const writeToStream = () => {
-        let cont = () => {
+        const cont = () => {
           i++;
           if (i < buffers.length) {
             setImmediate(writeToStream);
           } else {
             stream.end();
           }
-        }
+        };
         const drained = stream.write(buffers[i]);
         if (drained) {
           cont();
         } else {
-          stream.once('drain', writeToStream);
+          stream.once('drain', cont);
         }
-      }
+      };
       writeToStream();
     });
     server.listen(0);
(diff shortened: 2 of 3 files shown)


mcollina:
LGTM with a test.

kjin:
@mcollina Thanks for looking at this. I've added a test which sets up servers that send a large amount of data to clients with varying window sizes.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+        const drained = stream.write(buffers[i], 'buffer', resolve);
+        if (!drained) {
+          stream.once('drain', resolve);
+        }
+      }));
+    }
+    p = p.then(() => {
+      stream.end();
+    });
+  });
```
mcollina: this block is not really idiomatic stream code, also it allocates a lot of functions and promises that are not needed. Within a function, you can iterate over `buffers` as long as `write` returns `true`, when `false`, you can setup a `once('drain')` handler to restart from where it stopped.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+  const resultLength =
+    buffers.reduce((prev, buf) => prev + buf.length, 0);
+  const result = Buffer.alloc(resultLength);
+  let offset = 0;
+  for (let i = 0; i < buffers.length; i++) {
+    buffers[i].copy(result, offset);
+    offset += buffers[i].length;
+  }
+  return result;
+}
```
mcollina: There is `Buffer.concat` for this exact purpose.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+  undefined, // use default window size which is (1 << 16) - 1
+];
+
+// Call `run` on each element in the cartesian product of buffersList and
+// initialWindowSizeList.
+for (const buffers of buffersList) {
+  for (const initialWindowSize of initialWindowSizeList) {
+    run(buffers, initialWindowSize);
+  }
+}
```
mcollina: it will be better for debuggability to run those one at a time.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+        const req = client.request({
+          ':method': 'GET',
+          ':path': '/'
+        });
+        let actualBuffer = Buffer.alloc(expectedBuffer.length);
+        let offset = 0;
+        req.on('data', data => {
+          data.copy(actualBuffer, offset);
+          offset += data.length;
+        });
```
mcollina: I think it's better to store all the chunks in an array, and then call `Buffer.concat()`.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+        req.on('end', common.mustCall(() => {
+          assert.strictEqual(offset, actualBuffer.length);
+          assert.strictEqual(Buffer.compare(actualBuffer, expectedBuffer), 0);
+          // shut down
+          client.socket.destroy();
```
mcollina: does `client.destroy()` has bugs and does not work? https://github.com/nodejs/http2/blob/master/lib/internal/http2/core.js#L435

test/parallel/test-http2-client-window-size.js
jasnell: Since this is a test, you could also make use of `/internals/streams/BufferList.js` .. it's internal code but if you include `// Flags: --expose-internals` at the top of the test it will be accessible.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+      }));
+  }));
+}
+
+const bufferValueRange = [0, 1, 2, 3];
+const buffersList = [
+  bufferValueRange.map(a => Buffer.alloc(1 << 4).fill(a)),
+  bufferValueRange.map(a => Buffer.alloc((1 << 8) - 1).fill(a)),
+  bufferValueRange.map(a => Buffer.alloc(1 << 8).fill(a)),
+  bufferValueRange.map(a => Buffer.alloc(1 << 17).fill(a))
```
jasnell: These can be simplified to `Buffer.alloc(n, a)` where `n` is the size. The second argument of `Buffer.alloc()` is the fill value.

test/parallel/test-http2-client-window-size.js
```diff
@@ -0,0 +1,104 @@
...
+
+// Given a list of buffers and an initial window size, have a server write
+// each buffer to the HTTP2 Writable stream, and let the client verify that
+// all of the bytes were sent correctly
+function run(buffers, initialWindowSize) {
+  const expectedBuffer = concatBuffers(buffers);
+
+  const server = h2.createServer();
+  const a = [];
+  server.on('stream', (stream, headers, flags) => {
```
jasnell: Since `headers` and `flags` are not used, they can be omitted

jasnell (changes requested):
Would like to see @mcollina's comments addressed + a couple more.

test/parallel/test-http2-client-window-size.js
kjin: Sure thing -- though, on that note, this test takes a few seconds to run, and running these sequentially would probably take even longer. Do you have any suggestions on how this could be cut down? And whether it's necessary to do so?

test/parallel/test-http2-client-window-size.js
kjin: Ah, I forgot to report this as an issue -- but `client.destroy` doesn't destroy the socket. It seems like `socket.destroy` calls `destroy` on the associated session, and I would have expected to be the other way around. In any case, I'll open an issue on this (edit: #62).

test/parallel/test-http2-client-window-size.js
kjin: I think `Buffer.concat` should be fine for what I have here.

kjin:
@jasnell @mcollina I've uploaded changes to address your comments.

Something I should mention is that when the test cases within this file are run sequentially, the test takes a long time (estimating around a minute). Given this, would it be wise to narrow the test cases  to just a few (or revert to having them parallelized)?

jasnell:
It should be both. The session and socket should be bound such that if
either closed, the other is also closed, regardless of which close() method
is called.

jasnell (approved):

jasnell:
Running the test locally, it appears to hang and not exit. Will have to investigate what is happening in a bit.

jasnell:
@kjin ... yeah, this is definitely hanging up somewhere in the test. A very quick glance through does not reveal exactly where the issue is. I'll try to work through it here soon but if you can take a look at see if you can find it, that would be helpful also.

kjin:
@jasnell Thanks for looking into this. I think the test hangs because of the case where the server is sending `(1<<17)*4` bytes to the client. I am investigating now on why this is happening -- it used to not be an issue before.

But also, if the time it takes to send this many bytes is prohibitively high and there's no way to make it faster, then what would you suggest when it comes to ensuring that a large number of bytes make it through? I wasn't 100% sure having a test case with this many bytes was necessary, and I was looking for feedback on that.

kjin:
I realized that the hanging (+ perception of slowness that I mentioned earlier) comes from a bug I introduced in the test. I've uploaded the necessary fix and linted the changed code.

jasnell:
I'll test this again shortly

test/parallel/test-http2-window-size.js
```diff
@@ -0,0 +1,98 @@
+'use strict';
+
+// This test ensures that servers are able to send data independent of window
+// size.
+
```
jasnell: There is a chance, on some smaller / iot platforms this test could be problematic because of the large buffer allocations. May be a  good idea to include a TODO note in here to test on those machines... and if it ends up being too much of a burden, we might want to move this test out of parallel.

jasnell (approved):
LGTM

test/parallel/test-http2-window-size.js
kjin: Ok, added. Once again, thanks for taking the time to look at this.

jasnell:
Landed! Thank you!

PR #61, opened 2017-05-02
http2: specify default TLS options for http2 client connection.

### LINKED ISSUE #59: a better default configuration for TLS connection of http2
It seems the TLS socket is initialized as `tls.connect(port, host, options)` in `http2.connect`. Some existing HTTP2 servers require some TLS options to set up the connection properly. More specifically:
- servername (for Server Name Indication)
- ALPN / NPN, specifying 'h2'

I believe those should be set by defaults, then passing a URL string will just work.

Also, maybe it's better to use `secureConnect` event instead of `connect` event for the TLS sockets for setting up a session -- so that it won't go into the session if some TLS setup failure happens.

jmuk:
fixes: https://github.com/nodejs/http2/issues/59

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1084,10 +1084,13 @@ function initializeOptions(options) {
   return options;
 }
 
-function initializeTLSOptions(options) {
+function initializeTLSOptions(options, servername) {
   options = initializeOptions(options);
   options.ALPNProtocols = ['hc', 'h2'];
   options.NPNProtocols = ['hc', 'h2'];
+  if (servername !== undefined) {
+    options.servername = servername;
+  }
   return options;
 }
 
@@ -1235,12 +1238,7 @@ function connect(authority, options, listener) {
       socket = net.connect(port, host);
       break;
     case 'https:':
-      options = util._extend({
-        servername: host,
-        ALPNProtocols: ['h2'],
-        NPNProtocols: ['h2']
-      }, options);
-      socket = tls.connect(port, host, options);
+      socket = tls.connect(port, host, initializeTLSOptions(options, host));
       break;
     default:
       throw new TypeError(`protocol "${protocol}" in unsupported.`);
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1235,6 +1235,11 @@ function connect(authority, options, listener) {
       socket = net.connect(port, host);
       break;
     case 'https:':
+      options = util._extend({
+        servername: host,
+        ALPNProtocols: ['h2'],
+        NPNProtocols: ['h2']
+      }, options);
       socket = tls.connect(port, host, options);
       break;
     default:
--- a/test/parallel/test-http2-create-client-secure-session.js
+++ b/test/parallel/test-http2-create-client-secure-session.js
@@ -18,10 +18,9 @@ function loadKey(keyname) {
 }
 
 const server = h2.createSecureServer({cert, key});
-const count = 10;
 
 // we use the lower-level API here
-server.on('stream', common.mustCall(onStream, count));
+server.on('stream', common.mustCall(onStream));
 
 function onStream(stream) {
   stream.respond({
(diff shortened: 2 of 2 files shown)


lib/internal/http2/core.js
```diff
@@ -1235,6 +1235,11 @@ function connect(authority, options, listener) {
       socket = net.connect(port, host);
       break;
     case 'https:':
+      options = util._extend({
+        servername: host,
+        ALPNProtocols: ['h2'],
+        NPNProtocols: ['h2']
+      }, options);
```
mcollina: I would prefer `Object.assign`

mcollina (changes requested):
can you also add a unit test for this?

jasnell:
Had a different approach in mind for fixing this: 

```js
diff --git a/lib/internal/http2/core.js b/lib/internal/http2/core.js
index 6afc87e6f7..052513289f 100755
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1077,6 +1077,7 @@ function initializeOptions(options) {
   options = options || {};
   if (typeof options !== 'object')
     throw new TypeError('options must be an object');
+  options = Object.create(options);
   options.allowHalfOpen = true;
   options.settings = options.settings || {};
   if (typeof options.settings !== 'object')
@@ -1084,10 +1085,12 @@ function initializeOptions(options) {
   return options;
 }

-function initializeTLSOptions(options) {
+function initializeTLSOptions(options, host) {
   options = initializeOptions(options);
   options.ALPNProtocols = ['hc', 'h2'];
   options.NPNProtocols = ['hc', 'h2'];
+  if (host !== undefined)
+    options.servername = host;
   return options;
 }

@@ -1235,7 +1238,7 @@ function connect(authority, options, listener) {
       socket = net.connect(port, host);
       break;
     case 'https:':
-      socket = tls.connect(port, host, options);
+      socket = tls.connect(port, host, initializeTLSOptions(options, host));
       break;
     default:
       throw new TypeError(`protocol "${protocol}" in unsupported.`);
james@ubuntu:~/node/http2-jasnell$
```

This allows us to have a central place for setting the TLS options for both server and client.

jmuk:
Thank you for the review!  Added a test case, and fixed the client code for your latter suggestion.

jmuk:
By the way -- where does the ALPN name `'hc'` come from? I can't find the document referring that value -- http2 spec would suggest to use `h2` only for TLS context, I suppose (see https://http2.github.io/http2-spec/#rfc.section.3.1).

jasnell:
I think it's an old pre-standard value. One of the browser impls was using it. Whether or not we need to keep it remains to be seen

test/parallel/test-http2-create-client-secure-session.js
```diff
@@ -0,0 +1,73 @@
...
+        const jsonData = JSON.parse(data);
+        assert.strictEqual(jsonData.servername, 'localhost');
+        assert(jsonData.alpnProtocol === 'h2' || jsonData.alpnProtocol === 'hc');
+        if (--expected === 0) {
+          server.close();
+          client.socket.destroy();
+        }
+      }));
+      req.end();
+    }
```
mcollina: why are you repeating this test 10 times?

jmuk: I borrowed the structure from `test/parallel/test-http2-create-client-session.js` which also repeats 10 times (https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-create-client-session.js), I think it's better to have the same structure here. Should I change here, or both?

jasnell: the calling ten times in the other test is an artifact from trying to trace down a bug I was working through. I actually need to refactor that test to pull that back out :-)

jmuk:
Removed the count, thank you for the background info.

jasnell (approved):

jasnell:
@mcollina ... ping

mcollina (approved):

mcollina:
LGTM

jasnell:
Landed!

PR #65, opened 2017-05-04
http2: fix test-http2-binding after round 11

mcollina:
As titled, there was a broken test.

### DIFF
--- a/test/parallel/test-http2-binding.js
+++ b/test/parallel/test-http2-binding.js
@@ -13,7 +13,7 @@ assert(http2.SessionShutdownWrap);
 assert.strictEqual(typeof http2.Http2Session, 'function');
 assert.strictEqual(typeof http2.SessionShutdownWrap, 'function');
 
-const settings = http2.getDefaultSettings({});
+const settings = require('http2').getDefaultSettings();
 assert.strictEqual(settings.headerTableSize, 4096);
 assert.strictEqual(settings.enablePush, true);
 assert.strictEqual(settings.initialWindowSize, 65535);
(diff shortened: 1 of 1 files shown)


danbev (approved):
All tests pass for me now.

test/parallel/test-http2-binding.js
```diff
@@ -25,10 +25,9 @@ assert.strictEqual(http2.nghttp2ErrorString(-517),
 const check = Buffer.from([0x00, 0x01, 0x00, 0x00, 0x10, 0x00, 0x00, 0x04,
                            0x00, 0x00, 0xff, 0xff, 0x00, 0x05, 0x00, 0x00,
                            0x40, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x01]);
-const val = http2.packSettings(http2.getDefaultSettings({}));
+const val = http2.packSettings(require('http2').getDefaultSettings());
 assert.deepStrictEqual(val, check);
 
-assert.doesNotThrow(() => assert(require('http2').getDefaultSettings()));
```
robertkowalski, line 31: just curious: is it that after a `packSettings` call `getDefaultSettings()` should throw?

test/parallel/test-http2-binding.js
```diff
@@ -13,7 +13,7 @@ assert(http2.SessionShutdownWrap);
 assert.strictEqual(typeof http2.Http2Session, 'function');
 assert.strictEqual(typeof http2.SessionShutdownWrap, 'function');
 
-const settings = http2.getDefaultSettings({});
+const settings = require('http2').getDefaultSettings();
```
robertkowalski, line 16: why is require used here, I believe the module cache delivers ithe module anyway from the cache?

jasnell (approved):

robertkowalski:
makes all sense now, thanks for the explanations!

test/parallel/test-http2-binding.js
jasnell: no, those are separate operations. `packSettings()` returns a base64 encoded string with the settings, `getDefaultSettings()` returns the default settings object.

jasnell:
Note that this test could likely use some general refactoring. When I wrote it initially the goal was just to do a quick rudimentary test that things we generally ok. I was not writing it for quality.

jasnell:
Landed!

PR #80, opened 2017-05-05
http2: an accumulation of another round of edits

jasnell:
I will be updating this incrementally over the next few days. There are multiple general edits in here.


lib/internal/http2/core.js
```diff
@@ -200,33 +231,89 @@ function onSessionShutdownComplete(status, wrap) {
...
+  if (options.silent === undefined)
+    options.silent = false;
+  else if (typeof options.silent !== 'boolean') {
+    const err = new errors.RangeError('ERR_INVALID_OPT_VALUE',
+                                      'silent',
+                                      options.silent);
+    Error.captureStackTrace(err, validatePriorityOptions);
+    throw err;
+  }
+}
```
mcollina, line 321: this is not really readable, I would prefer a blank line between the if blocks, and braces everywhere.

lib/internal/http2/core.js
```diff
@@ -332,15 +424,15 @@ class Http2Session extends EventEmitter {
     // streams, these may all need to be made aware of changes
     // in state that occur -- such as when the associated socket
     // is closed. To do so, we need to set the max listener count
-    // to something more reasonable.
+    // to something more reasonable to avoid the memory leak warning.
```
mcollina: avoiding a warning is not newbie friendly, as it encourages a good practice. The actual reason is that we can have an extremely high number of streams multiplexed over the session.

lib/internal/http2/core.js
```diff
@@ -355,7 +447,7 @@ class Http2Session extends EventEmitter {
 
   get destroyed() {
     const state = this[kState];
-    return Boolean(state.destroyed);
+    return !!state.destroyed;
```
mcollina: do we need this check at all? I think we should always have a `destroyed` property in the state.

lib/internal/http2/core.js
```diff
@@ -676,7 +804,7 @@ class Http2Stream extends Duplex {
     req.callback = cb;
     req.oncomplete = afterDoStreamWrite;
     req.async = false;
-    const enc = data instanceof Buffer ? 'buffer' : encoding;
+    const enc = Buffer.isBuffer(data) ? 'buffer' : encoding;
```
mcollina: is this needed? `encoding` should already contain this info.

mcollina (approved):
Some nits, LGTM after those.

lib/internal/http2/core.js
jasnell: likely a good idea

lib/internal/http2/core.js
jasnell: yep, you're right :-)

jasnell:
@mcollina ... updated

mcollina (approved):
LGTM

jasnell:
added some linting fixes and a fix for server-rst-stream test

jasnell:
Rebased and updated. PTAL

jasnell:
CI: https://ci.nodejs.org/job/node-test-pull-request/8049/

(first time running the http2 branch on our CI... fingers crossed)

jasnell:
CI shows a handful of failures that would need to be looked at but nothing specific to this PR. Landing

jasnell:
Landed!

PR #95, opened 2017-05-09
http2: add server push event for client

yosuke-furukawa:
http2 server push does not throw any event for clients, and if we use pushStream method, stream can not be found, so throw NPE.

this pull request emit `push` event for client and avoid the NPE. 

 

##### Checklist

http2, test

### DIFF
--- a/test/parallel/test-http2-server-push-stream.js
+++ b/test/parallel/test-http2-server-push-stream.js
@@ -1,0 +1,52 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const http2 = require('http2');
+
+const server = http2.createServer();
+server.on('stream', common.mustCall((stream, headers, flags) => {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
LGTM

jasnell (approved):

test/parallel/test-http2-server-push-stream.js
```diff
@@ -0,0 +1,52 @@
...
+
+  let data = '';
+
+  req.on('data', common.mustCall((d) => data += d));
+  req.on('end', common.mustCall(() => {
+    assert.strictEqual(data, 'test');
+    server.close();
+    client.destroy();
+  }));
+  req.end();
```
robertkowalski, line 51: i think you don't need to call end, as the server ends the repsonse

mcollina:
Landed as 14971f5

PR #100, opened 2017-05-11
src: add assertion to check cb is not null

danbev:
Currently nghttp2_free_headers_list uses the cb pointer but does not
check it for null first like other functions do. Wanted to bring this up
just in case it was simply overlooked.

##### Checklist

src


mcollina (approved):
LGTM

cjihrig (approved):

jasnell (approved):

mcollina:
Landed as a11cf95094609fa0802c005924a3f837a20d1300

PR #102, opened 2017-05-12
Docs!

jasnell:
Initial docs PR .. still lots of detail to fill in.

### DIFF
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -1,0 +1,727 @@
+# HTTP2
+
+> Stability: 1 - Experimental
+
+The `http2` module provides an implementation of the [HTTP/2][] protocol. It
+can be accessed using:
+
+```js
+const http2 = require('http2');
+```
+
+## Core API
+
+The Core API provides a low-level interface designed specifically around
+support for HTTP/2 protocol features. It is specifically *not* designed for
+compatibility with the existing [HTTP/1][] module API.
+
+The following illustrates a simple, plain-text HTTP/2 server:
+
+```js
+const http2 = require('http2');
+
+// Create a plain-text HTTP/2 server
+const server = http2.createServer();
+
+server.on('stream', (stream, headers) => {
+  stream.respond({
+    'content-type': 'text/html',
+    ':status': 200
+  });
+  stream.end('<h1>Hello World</h1>');
+});
+
+server.listen(80);
+```
+
+The following illustrates an HTTP/2 client:
+
+```js
+const http2 = require('http2');
+
+const client = http2.connect('http://localhost:80');
+
+const req = client.request({ ':path': '/'});
+
+req.on('response', (headers) => {
+  console.log(headers[':status']);
+  console.log(headers['date']);
+});
+
+let data = '';
+req.setEncoding('utf8');
+req.on('data', (d) => data += d);
+req.on('end', () => client.destroy());
+req.end();
+```
+
+### Class: Http2Session
+
+* Extends: {EventEmitter}
+
+Instances of the `http2.Http2Session` class represent an active communications
+session between an HTTP/2 client and server. Instances of this class are *not*
+intended to be constructed directly by user code.
+
+Every `Http2Session` instance is associated with exactly one [`net.Socket`][] or
+[`tls.TLSSocket`][] when it is created. When either the `Socket` or the
+`Http2Session` are destroyed, both will be destroyed.
+
+Each `Http2Session` instance will exhibit slightly different behaviors
+depending on whether it is operating as a server or a client. The
+`http2session.type` property can be used to determine the mode in which an
+`Http2Session` is operating. On the server side, user code should rarely
+have occasion to work with the `Http2Session` object directly, with most
+actions typically taken through interactions with either the `Http2Server` or
+`Http2Stream` objects.
+
+#### Event: 'close'
+
+The `'close'` event is emitted once the `Http2Session` has been terminated.
+
+#### Event: 'connect'
+
+The `'connect'` event is emitted once the `Http2Session` has been successfully
+connected to the remote peer and communication may begin.
+
+*Note*: User code will typically not listen for this event directly.
+
+#### Event: 'error'
+
+(TODO: fill in detail)
+
+#### Event: 'selectPadding'
+
+(TODO: fill in detail)
+
+#### Event: 'stream'
+
+The `'stream'` event is emitted when a new `Http2Stream` is created. When
+invoked, the handler function will receive a reference to the `Http2Stream`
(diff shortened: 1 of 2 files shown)


doc/api/http2.md
```diff
@@ -0,0 +1,438 @@
...
+
+Immediately terminates the `Http2Session` and the associated `net.Socket` or
+`tls.TLSSocket`.
+
+#### http2session.destroyed
+
+* Value: {boolean}
+
+Will be `true` if this `Http2Session` instance has been destroyed and must no
+longer be used, otherwise `false.
```
joyeecheung: This is missing a backtick at the end so the following text are not highlighted correctly..

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+### Class: Http2Session
+
+* Extends: {EventEmitter}
+
+Instances of the `http2.Http2Session` class represent an active communications
+session between an HTTP/2 client and server. Instances of this class are *not*
+intended to be constructed directly by user code.
+
+Every `Http2Session` object is associated with exactly one [`net.Socket`][] or
+[`tls.TLSSocket`][]. Upon creation of the `Http2Session` instance, the `Socket`
```
cjihrig: Could you rephrase the sentence that starts with "Upon". It reads a little awkwardly IMO.

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+will be associated, with the lifecycle of each being bound to the other. When
+either is destroyed, both will be destroyed.
+
+#### Event: `'close'`
```
cjihrig: Looking at the existing http docs, it doesn't look like backticks are used in the event headers (but the single quotes are).

cjihrig (commented):
All of the APIs need the `<!-- YAML` tags.

Did you mean to change the mode on doc/api/index.md?

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+    ':status': 200
+  });
+  stream.end('<h1>Hello World</h1>');
+});
+
+server.listen(80);
+```
+
+The following illustrates an HTTP/2 client:
+```
```
kjin: The code example here doesn't have `js` highlighting applied to it.

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+* `callback` {Function}
+
+(TODO: fill in detail)
+
+#### http2session.shutdown(options[, callback])
+
+* `options` {Object}
+  * `graceful` {boolean} `true` to attempt a polite shutdown of the
+    `Http2Session`.
+  * `immediate` (boolean) `true` to force the shutdown to occur immediately,
```
kjin: Nit - should this be {boolean}?

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+});
+
+let data = '';
+req.setEncoding('utf8');
+req.on('data', (d) => data += d);
+req.on('end', () => client.destroy());
+req.end();
+```
+
+### Class: Http2Session
```
kjin, line 58: It seems like on the server side one never directly interacts with `Http2Session` objects. Do you think it would be helpful to make a note of that?

doc/api/http2.md
```diff
@@ -0,0 +1,692 @@
...
+The `http2session.type` will be equal to
+`http2.constants.NGHTTP2_SESSION_SERVER` if this `Http2Session` instance is a
+server, and `http2.constants.NGHTTP2_SESSION_CLIENT` if the instance is a
+client.
+
+### Class: Http2Stream
+
+* Extends: {Duplex}
+
+Instances of the `Http2Stream` class represent a bidirectional HTTP/2
```
kjin: Making 'Instances' singular (e.g. 'Each instance') might improve the readability of this sentence. Also, would it be within the scope of this section to give a brief overview of when `Http2Stream` objects are created? For example, to mention that on the client, it's returned by `Http2Session::request`, while on the server, it's passed through the `'stream'` event.

In addition, nit in the last sentence -- 'its lifetime'.

doc/api/http2.md
```diff
@@ -0,0 +1,727 @@
+# HTTP2
+
+> Stability: 1 - Experimental
+
+The `http2` module provides an implementation of the [HTTP/2][] protocol. It
+can be accessed using:
+
+```js
+const http2 = require('http2');
+```
```
mcollina, line 10: if we are not shipping this in `8.0.0`, we need to ship it behind a flag, as we will be shadowing the `'http2'` module otherwise. Maybe we might do something different here.

doc/api/http2.md
```diff
@@ -0,0 +1,727 @@
...
+
+The `'push'` event is emitted when response headers for a Server Push stream
+are received. The listener callback is passed the [Headers Object][] and flags
+associated with the headers.
+
+```js
+stream.on('push', (headers, flags) => {
+  // ...
+});
+```
```
mcollina: we need an example here.

doc/api/http2.md
```diff
@@ -0,0 +1,727 @@
...
+
+The `'request'` event is emitted when a block of headers associated with an
+HTTP request is received. The listener callback is passed the [Headers Object][]
+and flags associated with the headers.
+
+```js
+stream.on('request', (headers, flags) => {
+  // ...
+});
+```
```
mcollina: and here

doc/api/http2.md
```diff
@@ -0,0 +1,727 @@
...
+```js
+stream.on('response', (headers, flags) => {
+  // ...
+});
+```
+
+This is emitted only when `http2session.type` is equal to
+`http2.constants.NGHTTP_SESSION_CLIENT`.
+
+#### Event: `'streamClosed'
```
mcollina: missing `

doc/api/http2.md
jasnell: If it's behind a flag (thanks for the reminder btw) then I don't believe that we'll need to do anything different here. We'll just need to document the flag for the time being. I'll work on getting the flag added this week

doc/api/http2.md
jasnell: agreed.... but let's add that separately once the push stuff is a bit more settled.

jasnell: In fact, we can keep this PR open for a bit and anyone who would like to push commits against it to fill in examples is more than welcome :-D

mcollina:
I think we should start merging this, and then we can improve it along the way.

jasnell:
@mcollina ... does that count as sign off?

mcollina (approved):
LGTM

jasnell:
Landed!

PR #109, opened 2017-05-13
http2: more efficient session and stream state, and local/remote settings

jasnell:
* Use more efficient mechanism for retrieving stream and session state and local/remote settings
* Add tests
* Fix segfaults
* Fixing accidental file mode changes


lib/internal/http2/core.js
```diff
@@ -74,6 +74,34 @@ const {
   HTTP_STATUS_SWITCHING_PROTOCOLS
 } = constants;
 
+const defaultSettings = new Uint32Array(binding.defaultSettingsArrayBuffer);
+const settingsBuffer = new Uint32Array(binding.settingsArrayBuffer);
+const sessionState = new Float64Array(binding.sessionStateArrayBuffer);
+const streamState = new Float64Array(binding.streamStateArrayBuffer);
```
mcollina: can you please add a comment here to describe how those are used? They are a bit unclear.

Also, I might worry a bit about those `Float64`, as it seems we are casting integers there.

mcollina (approved):
LGTM with some nits

jasnell:
Landed with comments added

Issue #69, opened 2017-05-04
http2: prioritization api / support


jasnell:
There are several things needed here:

1. Method for setting the priority (this is already available)
2. Event when a priority change request has been received.
3. Event when the priority of a stream has actually changed.

jasnell:
Items 1 and 2 are done. Item 3 likely won't be possible. There's no notification from nghttp when it has applied any changes to priority. We'll have to track that on our own or get more sophisticated later.

PR #110, opened 2017-05-14
http2: add priority advisement event

jasnell:
Trigger a 'priority' event on the session, stream, and server when a priority frame is received.

Refs: https://github.com/nodejs/http2/issues/69

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -234,6 +234,24 @@ function onSettings() {
   owner[kRemoteSettings] = undefined;
 }
 
+// If the stream exists, an attempt will be made to emit an event
+// on the stream object itself. Otherwise, forward it on to the
+// session (which may, in turn, forward it on to the server)
+function onPriority(id, parent, weight, exclusive) {
+  debug(`priority advisement for stream ${id}: \n` +
+        `  parent: ${parent},\n  weight: ${weight},\n` +
+        `  exclusive: ${exclusive}`);
+  timers._unrefActive(this);
+  const owner = this[kOwner];
+  const state = owner[kState];
+  const streams = state.streams;
+  const stream = streams.get(id);
+  if (stream === undefined ||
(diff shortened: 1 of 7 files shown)


mcollina (changes requested):
Shouldn't priority have a callback attached?

jasnell:
attached where? not sure I follow

mcollina:
Sorry, I meant req.priority(obj, callback)

jasnell:
No, at least not right now. Calling `req.priority()` tells nghttp to send a priority frame, which is done on the next tick of the event loop when nghttp is prompted to send whatever frames it has queued up. Currently, that's a fire and forget mechanism. We can make it more intelligent or introduce a callback later if need arises, but for the time being, there does not appear to be a need for it.

mcollina:
@jasnell can you add a comment on this? From a consistency perspective, a synchronous API that does something on the next event loop tick is not consistent. I agree that there is no need for having a callback right now, but we have to document it properly.

mcollina (approved):
LGTM

yosuke-furukawa:
LGTM

jasnell:
Yes, I will add a comment... however, it's the same situation for most of the operations that cause nghttp to queue data up to be sent. Specifically, the methods either queue things up in nghttp's own internal frame buffer or in our link lists... then on the next tick, nghttp is prompted to begin sending the buffered data, which drains it's own internal buffer and our linked list. The data writes are the only thing that currently have callbacks attached to them. I'll make a note to look into how feasible it would be to add callbacks to everything.

mcollina:
@jasnell most of the other APIs are behind streams interactions, which makes them easy to understand from a user point of view. At least, they know they will be asynchronous. `priority` on the other end it is async, but it looks sync. I think we can move forward here, just add a comment so that we know it is something to document well.

jasnell:
Landed with note added

PR #111, opened 2017-05-15
http2: fixed .finished property on req/res compat

sebdeckers:
As per Node.js HTTP API docs:

1. Removed `request.finished` property; and
1. Added and fixed [`response.finished`](https://nodejs.org/api/http.html#http_response_finished) property.

##### Checklist

- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-guidelines)

http2

### DIFF
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -3,9 +3,7 @@
 const Stream = require('stream');
 const Readable = Stream.Readable;
 const binding = process.binding('http2');
-const {
-  isIllegalConnectionSpecificHeader,
-} = require('internal/http2/util');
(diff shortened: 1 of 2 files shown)


sebdeckers:
Note that this defaults the `finished` property to `false` in case there is no underlying stream. The idea is that this would happen for streams that have ended and were cleaned-up. However for streams that have not yet been created this could be misleading. Though I'm not sure if it's possible, nor valid, to arrive at such a state.

mcollina (approved):
LGTM

cjihrig (approved):

jasnell (approved):

jasnell:
Landed!

PR #112, opened 2017-05-15
http2: writeHead optional statusMessage

sebdeckers:
Adds backwards compatible support for the optional `statusMessage` argument to [`response.writeHead`](https://nodejs.org/api/http.html#http_response_writehead_statuscode_statusmessage_headers).

The status message is [not used in HTTP/2](http://httpwg.org/specs/rfc7540.html#HttpResponse). This optional argument is only exposed to stay synched with the `require('http')` API.

##### Checklist

- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-guidelines)

http2

### DIFF
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -386,7 +386,12 @@ class Http2ServerResponse extends Stream {
     this.writeHead(statusCode);
   }
 
-  writeHead(statusCode, headers) {
+  // HTTP/2 has no status message support (RFC7540 8.1.2.4), so
+  // statusMessage is ignored and only allowed for backwards compatibility.
+  writeHead(statusCode, statusMessage, headers) {
+    if (headers === undefined && typeof statusMessage === 'object') {
+      headers = statusMessage;
+    }
     var state = this[kState];
     if (state.headersSent === true)
       return;
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -389,14 +389,12 @@ class Http2ServerResponse extends Stream {
   }
 
   writeHead(statusCode, statusMessage, headers) {
-    if (typeof statusMessage === 'string') {
-      if (statusMessageWarned === false) {
-        process.emitWarning(
-          'Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)',
-          'UnsupportedWarning'
-        );
-        statusMessageWarned = true;
-      }
+    if (typeof statusMessage === 'string' && statusMessageWarned === false) {
+      process.emitWarning(
+        'Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)',
+        'UnsupportedWarning'
+      );
+      statusMessageWarned = true;
     }
     if (headers === undefined && typeof statusMessage === 'object') {
       headers = statusMessage;
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -17,6 +17,8 @@ const kResponse = Symbol('response');
 const kHeaders = Symbol('headers');
 const kTrailers = Symbol('trailers');
 
+let statusMessageWarned = false;
+
 // Defines and implements an API compatibility layer on top of the core
 // HTTP/2 implementation, intended to provide an interface that is as
 // close as possible to the current require('http') API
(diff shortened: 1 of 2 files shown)


lib/internal/http2/compat.js
```diff
@@ -386,7 +386,10 @@ class Http2ServerResponse extends Stream {
     this.writeHead(statusCode);
   }
 
-  writeHead(statusCode, headers) {
+  writeHead(statusCode, statusMessage, headers) {
+    if (headers === undefined && typeof statusMessage === 'object') {
+      headers = statusMessage;
+    }
```
mcollina: would you mind adding a comment before `writeHead`, saying that `statusMessage` is ignored?

mcollina (changes requested):
LGTM with a nit

cjihrig (approved):

lib/internal/http2/compat.js
sebdeckers: Okay, updated. ✍️ (Feel free to rewrite the message. Not sure how verbose these comments ought to be.)

mcollina (approved):

jasnell (approved):
While I'm fine with the code change, the fact that the status message is just dropped on the floor is a bit irritating. Not sure there's anything else we can do tho

sebdeckers:
@jasnell FWIW even the HTTP/1 spec said the status message ("reason phrase") was to be ignored.

>    The reason-phrase element exists for the sole purpose of providing a
   textual description associated with the numeric status code, mostly
   out of deference to earlier Internet application protocols that were
   more frequently used with interactive text clients.  A client SHOULD
   ignore the reason-phrase content.

https://tools.ietf.org/html/rfc7230#section-3.1.2

jasnell:
Yes, I'm definitely quite aware of that. I also know that there are applications out there that have completely ignored that advice and make use of the status message for various purposes. Having it the value *appear* to work but having the data end up getting dropped on the floor seems problematic to me. I'd almost prefer to throw if it's set.

mcollina:
@jasnell the essence of the compat API is to provide some answer to those inconsistencies. I do not think we should throw, as it will negate part of the purpose. However, we might decide to be strict, and throw. If we decide to throw, we should add an option to disable that.

jasnell:
Maybe a process warning would be appropriate for the time being?

mcollina:
@jasnell that works.

sebdeckers:
@jasnell @mcollina I've changed the comment to a process warning. Not sure if the warning type `UnsupportedWarning` is appropriate.

lib/internal/http2/compat.js
```diff
@@ -386,7 +388,19 @@ class Http2ServerResponse extends Stream {
     this.writeHead(statusCode);
   }
 
-  writeHead(statusCode, headers) {
+  writeHead(statusCode, statusMessage, headers) {
+    if (typeof statusMessage === 'string') {
+      if (statusMessageWarned === false) {
```
jasnell: These can be combined...

```js
if (typeof statusMessage === 'string' && statusMessageWarned === false) {
  // ...
}
```

test/parallel/test-http2-compat-serverresponse-statusmessage.js
```diff
@@ -0,0 +1,47 @@
...
+// Http2ServerResponse.writeHead should accept an optional status message
+
+const server = h2.createServer();
+server.listen(0, common.mustCall(function() {
+  const port = server.address().port;
+  server.once('request', common.mustCall(function(request, response) {
+    const statusCode = 200;
+    const statusMessage = 'OK';
+    const headers = {'foo-bar': 'abc123'};
+    common.expectWarning(
```
jasnell: I'd move the call to `common.expectWarning()` up to the top level, outside of all the functions.

jasnell (approved):
LGTM with a couple of suggestions

test/parallel/test-http2-compat-serverresponse-statusmessage.js
sebdeckers: Hmm, but then it would also need to trap the warning emitted by `/lib/http2.js`, or do some `nextTick` deferral? Registering this assertion in the callback avoids that. (Not familiar at all with these APIs so please feel free to correct me.)

lib/internal/http2/compat.js
sebdeckers: Done 9bb976c

test/parallel/test-http2-compat-serverresponse-statusmessage.js
jasnell: If it's avoiding it it's because of a race condition that you're fortunate enough to be winning ;-) ... all `common.expectWarning()` does is add a listener to `process.on('warning')`. I'm not sure if it's currently set up to support listening for two types of warnings so it may be necessary to use your own `process.on('warning') handler instead (that can listen to either)

sebdeckers: `process.emitWarning` does:

```js
function doEmitWarning(warning) {
  return function() {
    process.emit('warning', warning);
  };
}
// ...
process.emitWarning = function(warning, type, code, ctor) {
  // ...
  process.nextTick(doEmitWarning(warning));
};
```
... which is why we'll always win that race condition. 😫

And `common.expectWarning` does support a map (type/description) to check multiple warnings, but it seems messy to include the second warning in this assertion. (Removing/editing the `lib/http2.js` warning shouldn't break this statusMessage test.)

So I'll add a custom handler, thanks for the suggestion!

jasnell:
Landed!

PR #115, opened 2017-05-15
http2: fix multivalued headers and cookies

jasnell:
* mapToHeaders function was not working correctly
* ensure that cookie headers are reassembled properly
* add test

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -66,6 +66,7 @@ const {
   HTTP2_HEADER_PATH,
   HTTP2_HEADER_SCHEME,
   HTTP2_HEADER_STATUS,
+  HTTP2_HEADER_COOKIE,
 
   HTTP_STATUS_CONTENT_RESET,
   HTTP_STATUS_OK,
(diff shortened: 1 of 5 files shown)


lib/internal/http2/core.js
```diff
@@ -156,6 +157,15 @@ function onSessionHeaders(id, cat, flags, headers) {
   const endOfStream = !!(flags & NGHTTP2_FLAG_END_STREAM);
   let stream = streams.get(id);
 
+  // https://tools.ietf.org/html/rfc7540#section-8.1.2.5
+  // "...If there are multiple Cookie header fields after decompression,
+  //  these MUST be concatenated into a single octet string using the
+  //  two-octet delimiter of 0x3B, 0x20 (the ASCII string "; ") before being
+  //  passed into a non-HTTP/2 context."
+  if (Array.isArray(headers[HTTP2_HEADER_COOKIE]))
```
sebdeckers, line 165: Is this considered a non-HTTP/2 context?

jasnell: That's a bit difficult to say because the spec does not really define that. It does mention passing the cookie value off to application server code with the implication that it counts as a 'non-HTTP/2 context' so I think this is appropriate here.

sebdeckers (approved):
LGTM

jasnell:
Landed!

PR #116, opened 2017-05-15
http2: add test for http methods

jasnell:
Adds a test to ensure that arbitrary http methods work.

##### Checklist

http2

### DIFF
--- a/test/parallel/test-http2-methods.js
+++ b/test/parallel/test-http2-methods.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const h2 = require('http2');
+
+const server = h2.createServer();
+
+const methods = [undefined, 'GET', 'POST', 'PATCH', 'FOO', 'A B C'];
+let expected = methods.length;
+
+// we use the lower-level API here
+server.on('stream', common.mustCall(onStream, expected));
+
+function onStream(stream, headers, flags) {
+  const method = headers[':method'];
+  assert.notStrictEqual(method, undefined);
+  assert(methods.includes(method), `method ${method} not included`);
+  stream.respond({
+    'content-type': 'text/html',
+    ':status': 200
+  });
+  stream.end('hello world');
+}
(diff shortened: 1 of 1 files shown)


test/parallel/test-http2-methods.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const h2 = require('http2');
+
+const server = h2.createServer();
+
+const methods = [undefined, 'GET', 'POST', 'PATCH', 'FOO', 'A B C'];
```
sebdeckers, line 9: Quote: http://httpwg.org/specs/rfc7231.html#method.overview

> The method token is case-sensitive because it might be used as a gateway to object-based systems with case-sensitive method names.

Add a lower- or mixed case method value.

> All general-purpose servers MUST support the methods GET and HEAD. All other methods are OPTIONAL.

Add HEAD as a method value.

Quote: http://httpwg.org/specs/rfc7230.html#rule.token.separators

> ```
>   token          = 1*tchar
> 
>   tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
>                  / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" 
>                  / DIGIT / ALPHA
>                  ; any VCHAR, except delimiters
> ```

Whitespace is not allowed so `A B C` would be invalid.

sebdeckers (changes requested):

test/parallel/test-http2-methods.js
jasnell: yep, the code does not currently do any validation, which is intentional for the time being. We'll need to go back in and add the necessary validation but we'll need to experiment with to find a method that doesn't kill performance.

sebdeckers (approved):

mcollina (approved):
LGTM

mcollina (approved):
LGTM

jasnell:
Landed!

Issue #118, opened 2017-05-15
http2: status code is string instead of number


robertkowalski:
The status code is a string now, not of type number any more.

Apart from the compat change, from a user perspective its a bit unhandy to get back the statuscode as a string, as you have to cast it before you make something like:

```js
if (statusCode >= 200) { // todo ... }
```

Example for failing, ported http1 test:

```js
'use strict';

const common = require('../common');
const assert = require('assert');

const http2 = require('http2');

const server = http2.createServer(function(request, response) {
  console.log('responding to ' + request.url);

  response.writeHead(200, {'Content-Type': 'text/plain'});
  response.write('1\n');
  response.write('');
  response.write('2\n');
  response.write('');
  response.end('3\n');

  this.close();
});

server.listen(0, common.mustCall(function() {
  const client = http2.connect(`http://localhost:${this.address().port}`);

  const headers = { ':path': '/' };
  const req = client.request(headers).setEncoding('ascii');

  let response = '';

  req.on('response', (headers) => {
    assert.strictEqual(200, headers[':status']);
  });

  req.on('data', function(chunk) {
    response += chunk;
  });

  req.on('end', common.mustCall(function(a,b,c) {
    assert.strictEqual('1\n2\n3\n', response);
  }));

  req.end();
}));

```

fails with:

```
assert.js:86
  throw new assert.AssertionError({
  ^
AssertionError: 200 === '200'
    at ClientHttp2Stream.req.on (/Users/robert/projects/http2/test/parallel/test-http2-write-empty-string.js:30:12)
    at emitTwo (events.js:125:13)
    at ClientHttp2Stream.emit (events.js:213:7)
    at Http2Session.onSessionHeaders [as onheaders] (internal/http2/core.js:194:12)
```

jasnell:
Yep... this would be a fairly easy fix if you'd like to give it a shot. There are three ways to approach it:

1. In the native later, if the header field name is `:key`, coerce the value to a number. This likely would be the much more difficult way.

2. In core.js, in the `onSessionHeaders` function, check for the existence of the `:status` header and coerce it's value to a number (e.g. `headers[HTTP2_HEADER_STATUS] |= 0`)

3. Coerce the value to a number in compat.js

Option 2 is likely the best choice.

robertkowalski:
Cool, I went for 2) and opened https://github.com/nodejs/http2/pull/120 -- I ran into an issue with closing the connection and its getting late here, maybe a bug?

[...]

PR #120, opened 2017-05-15
http2: statusCode should be a Number

robertkowalski:
Fixes https://github.com/nodejs/http2/issues/118

Issue I have right now: somehow the server / client hangs and does not end the test. req.resume() or req.end() are not called in the http1 implementation in https://github.com/nodejs/http2/blob/master/test/parallel/test-http-write-empty-string.js, so I'm puzzled.

##### Checklist

- [x] commit message follows [commit guidelines]
(https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-guidelines)

http2, test

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -101,6 +101,10 @@ function onSessionHeaders(id, cat, flags, headers) {
     headers[HTTP2_HEADER_COOKIE] =
       headers[HTTP2_HEADER_COOKIE].join('; ');
 
+  if (headers[HTTP2_HEADER_STATUS]) {
+    headers[HTTP2_HEADER_STATUS] |= 0;
+  }
+
   if (stream === undefined) {
     switch (owner.type) {
       case NGHTTP2_SESSION_SERVER:
--- a/test/parallel/test-http2-write-empty-string.js
+++ b/test/parallel/test-http2-write-empty-string.js
@@ -22,11 +22,11 @@ server.listen(0, common.mustCall(function() {
 
   let res = '';
 
-  req.on('response', (headers) => {
+  req.on('response', common.mustCall(function(headers) {
     assert.strictEqual(200, headers[':status']);
-  });
+  }));
 
-  req.on('data', function(chunk) {
+  req.on('data', (chunk) => {
     res += chunk;
   });
 
--- a/test/parallel/test-http2-write-empty-string.js
+++ b/test/parallel/test-http2-write-empty-string.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const http2 = require('http2');
+
+const server = http2.createServer(function(request, response) {
+  response.writeHead(200, {'Content-Type': 'text/plain'});
+  response.write('1\n');
+  response.write('');
+  response.write('2\n');
+  response.write('');
+  response.end('3\n');
+
+  this.close();
+});
+
+server.listen(0, common.mustCall(function() {
+  const client = http2.connect(`http://localhost:${this.address().port}`);
+  const headers = { ':path': '/' };
+  const req = client.request(headers).setEncoding('ascii');
+
+  let res = '';
+
+  req.on('response', (headers) => {
+    assert.strictEqual(200, headers[':status']);
+  });
+
+  req.on('data', function(chunk) {
+    res += chunk;
+  });
+
+  req.on('end', common.mustCall(function() {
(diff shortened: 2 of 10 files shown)


test/parallel/test-http2-write-empty-string.js
```diff
@@ -0,0 +1,40 @@
...
+  req.on('response', (headers) => {
+    assert.strictEqual(200, headers[':status']);
+  });
+
+  req.on('data', function(chunk) {
+    response += chunk;
+  });
+
+  req.on('end', common.mustCall(function(a,b,c) {
+    assert.strictEqual('1\n2\n3\n', response);
```
robertkowalski: end is called, but script keeps running

jasnell: Need to call `client.destroy()` and `server.close()` to shut things down.

robertkowalski: yeah i am calling `this.close` in L18

robertkowalski: `client.destroy` also doesn't work (in `req.on('end')`)...

test/parallel/test-http2-write-empty-string.js
```diff
@@ -0,0 +1,40 @@
...
+  console.log('responding to ' + request.url);
+
+  response.writeHead(200, {'Content-Type': 'text/plain'});
+  response.write('1\n');
+  response.write('');
+  response.write('2\n');
+  response.write('');
+  response.end('3\n');
+
+  this.close();
```
robertkowalski, line 15: `this.close` call

mcollina:
> Issue I have right now: somehow the server / client hangs and does not end the test. req.resume() or req.end() are not called in the http1 implementation in https://github.com/nodejs/http2/blob/master/test/parallel/test-http-write-empty-string.js, so I'm puzzled.

there is a `on('data')` event https://github.com/nodejs/http2/blob/master/test/parallel/test-http-write-empty-string.js#L47 on the response. `http.get()` automatically ends the requests.

test/parallel/test-http2-write-empty-string.js
mcollina: You need to close the client session.

Have you tried in moving `server.close()`  to `req.on('end')`, together with `server.destroy()`, as it's done everywhere else?

robertkowalski: will do. it wasn't clear to me what the expected behaviour of the new client is, but it takes shape now.

thank you for explaining and being so patient!

robertkowalski:
ready for review! :)

lib/internal/http2/core.js
```diff
@@ -101,6 +101,10 @@ function onSessionHeaders(id, cat, flags, headers) {
     headers[HTTP2_HEADER_COOKIE] =
       headers[HTTP2_HEADER_COOKIE].join('; ');
 
+  if (headers[':status']) {
```
cjihrig: Should this be `HTTP2_HEADER_STATUS`?

test/parallel/test-http2-write-empty-string.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+
```
cjihrig: Style nit, but there seems to be some randomly chosen blank lines in this test.

test/parallel/test-http2-write-empty-string.js
```diff
@@ -0,0 +1,43 @@
...
+const http2 = require('http2');
+
+const server = http2.createServer(function(request, response) {
+  console.log('responding to ' + request.url);
```
cjihrig: Can you remove the `console.log()`.

robertkowalski:
fixed & rebased

test/parallel/test-http2-write-empty-string.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+server.listen(0, common.mustCall(function() {
+  const client = http2.connect(`http://localhost:${this.address().port}`);
+  const headers = { ':path': '/' };
+  const req = client.request(headers).setEncoding('ascii');
+
+  let res = '';
+
+  req.on('response', (headers) => {
```
cjihrig: This should probably have a `common.mustCall()`.

cjihrig (approved):
LGTM with one last nit if @mcollina and @jasnell are good with it.

jasnell (approved):

mcollina:
Landed as 2061154

PR #123, opened 2017-05-16
http2: general code cleanups

jasnell:
A number of code cleanups that I've been accumulating

##### Checklist

http2

### DIFF
--- a/lib/http2.js
+++ b/lib/http2.js
@@ -6,17 +6,20 @@ process.emitWarning(
   'See https://github.com/nodejs/http2'
 );
 
-const core = require('internal/http2/core');
-//const compat = require('internal/http2/compat');
+const {
+  constants,
+  getDefaultSettings,
+  getPackedSettings,
+  createServer,
+  createSecureServer,
+  connect
+} = require('internal/http2/core');
 
-// Exports
 module.exports = {
-  constants: core.constants,
-  getDefaultSettings: core.getDefaultSettings,
-  getPackedSettings: core.getPackedSettings,
-  createServer: core.createServer,
-  createSecureServer: core.createSecureServer,
-  createServerSession: core.createServerSession,
-  connect: core.connect,
-  secureConnect: core.secureConnect
+  constants,
+  getDefaultSettings,
+  getPackedSettings,
+  createServer,
+  createSecureServer,
+  connect
 };
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1,45 +1,47 @@
 'use strict';
 
+const binding = process.binding('http2');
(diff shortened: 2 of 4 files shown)


lib/http2.js
```diff
@@ -6,17 +6,20 @@ process.emitWarning(
...
 module.exports = {
-  constants: core.constants,
-  getDefaultSettings: core.getDefaultSettings,
-  getPackedSettings: core.getPackedSettings,
-  createServer: core.createServer,
-  createSecureServer: core.createSecureServer,
-  createServerSession: core.createServerSession,
-  connect: core.connect,
-  secureConnect: core.secureConnect
+  constants,
```
sebdeckers, line 19: Why are `constants`, `getDefaultSettings`, and `getPackedSettings` exported?

lib/internal/http2/core.js
```diff
@@ -443,64 +325,101 @@ function setupHandle(session, socket, type, options, settings) {
...
 
+// Upon creation, the Http2Session takes ownership of the socket. The session
+// may not be ready to use immediately if the socket is not yet fully connected.
 class Http2Session extends EventEmitter {
-  constructor(type, options, socket, settings) {
+
+  // type     { number } either NGHTTP2_SESSION_SERVER or NGHTTP2_SESSION_CLIENT
+  // options  { Object }
+  // socket   { net.Socket | tls.TLSSocket }
+  // settings { Object }
```
sebdeckers: Remove settings line from comment

sebdeckers (commented):
Great stuff!

lib/http2.js
jasnell: They are all public APIs. `constants` includes constants that will be used by end users, `getDefaultSettings()` returns a pre-filled object that can be used by users to start with the default settings then modified, `getPackedSettings()` returns a buffer suitable for use with the `HTTP2-Settings` header field. These are specifically intended for end users to make use of.

sebdeckers: Thanks for clarifying. Haven't played with the settings much yet, still just doing basic stuff. 🤓

My concern with exposing the constants is that they contain many like `NGHTTP2_*`. Seems like something internal to the implementation, to be shielded from consumers.

mcollina (changes requested):
@jasnell have you checked if this effect throughput? I'm a bit concerned by all those `defineProperty` calls.

jasnell:
@mcollina ... yes, benchmark shows no significant change. In fact, setting the properties this way provides a slight perf improvement over getting at the data through getters.

mcollina (approved):
LGTM

jasnell:
Landed!

Issue #70, opened 2017-05-04
http2: padding support


jasnell:
How will this work?
Should we even support it in the first version?

PR #124, opened 2017-05-16
http2: improve padding callback

jasnell:
* Make padding callback more efficient using an array buffer,
* Use an options callback instead of an event

There's not much else here we really should do. Padding is not super important and we shouldn't get too fancy with it. This makes the callback fairly efficient for what it is.

Fixes: https://github.com/nodejs/http2/issues/70

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -45,6 +45,8 @@ const kProceed = Symbol('proceed');
 const kDefaultSocketTimeout = 2 * 60 * 1000;
 const kRenegTest = /TLS session renegotiation disabled for this socket/;
 
+const paddingBuffer = new Uint32Array(binding.paddingArrayBuffer);
+
 const {
   NGHTTP2_CANCEL,
   NGHTTP2_DEFAULT_WEIGHT,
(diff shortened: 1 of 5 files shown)


test/parallel/test-http2-padding-callback.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+server.listen(0);
+
+server.on('listening', common.mustCall(() => {
+  const client = h2.connect(`http://localhost:${server.address().port}`,
+                            options);
+
+  const req = client.request({ ':path': '/' });
+  req.on('response', common.mustCall());
```
sebdeckers, line 38: Is this test supposed to verify the presence of padding of received frames?

lib/internal/http2/core.js
```diff
@@ -442,7 +445,7 @@ function setupHandle(session, socket, type, options, settings) {
     handle.onstreamclose = onSessionStreamClose;
     handle.onerror = onSessionError;
     handle.onread = onSessionRead;
-    handle.ongetpadding = onSelectPadding;
+    handle.ongetpadding = onSelectPadding(options.selectPadding);
```
mcollina: we are allocating a closure for no reason if `options.selectPadding`  is undefined, as I expect it to be the "base case".

test/parallel/test-http2-padding-callback.js
jasnell: That's not particularly easy because nghttp2 automatically strips the padding. We cannot verify that the padding is correct because we never actually see it. The most we'd be able to do is verify the amount of padding reported in the frame header but we're currently not passing that information up to the js layer.

lib/internal/http2/core.js
jasnell: yeah, good point. if `options.selectPadding` is `undefined`, `handle.ongetpadding` should be `undefined`, in which case no padding would be applied.

mcollina (approved):
LGTM

jasnell:
Landed

PR #125, opened 2017-05-16
http2: ALPN fallback to HTTP/1.1

### LINKED ISSUE #122: http2: starting HTTP/2 with TLS ALPN
For anyone who may want to take this one on... 

----
[Starting HTTP/2 for "https" URIs](https://tools.ietf.org/html/rfc7540#section-3.3)

> A client that makes a request to an "https" URI uses TLS [TLS12] with the application-layer protocol negotiation (ALPN) extension [TLS-ALPN].

> HTTP/2 over TLS uses the "h2" protocol identifier.  The "h2c" protocol identifier MUST NOT be sent by a client or selected by a server; the "h2c" protocol identifier describes a protocol that does not use TLS.

> Once TLS negotiation is complete, both the client and the server MUST send a connection preface (Section 3.5).
----

Using the TLS ALPN mechanism, a server can theoretically be written to listen for both HTTP/1 and HTTP/2 requests, allowing the client to determine which to use based on the protocol identified via ALPN. To support this, a new TLSServer implementation would need to be created that selects either HTTP/1 or HTTP/2 based on the ALPN value. Since HTTP/1 and HTTP/2 will be two separate code paths, requiring different handlers, this could be a bit tricky to implement.

sebdeckers:
Fixes #122 

Well then, this was a lot easier than I thought it would be. 😌

- Deprecate NPN since this is not part of the HTTP/2 standard (and poorly documented AFAICT).
- Announce ALPN as `h2` or `http/1.1`, as per [IANA-registered protocol IDs](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids).
- For backwards compatibility pass the non-HTTP/2 socket to the HTTP/1.1 parser. This then emits `request` events on the `Http2SecureServer` instance. Analogous to the `https` module.

Note 1: ~~Some other tests are failing. Not sure if related to this patch.~~ TIL `--expose-internals`

Note 2: Only doing fallback to https clients. Not for plaintext http since ALPN is inherently TLS-only. Should use the [magic number to sniff plaintext http](https://tools.ietf.org/html/rfc7540#section-3.5).

##### Checklist

http2

### DIFF
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -611,6 +611,7 @@ server.listen(80);
   * `noRecvClientMagic` {boolean} (TODO: Add detail)
   * `paddingStrategy` {number} (TODO: Add detail)
   * `peerMaxConcurrentStreams` {number} (TODO: Add detail)
+  * `allowHTTP1` {boolean} Incoming client connections that do not support HTTP/2 will be downgraded to HTTP/1.x when set to `true`. The default value is `false`, which rejects non-HTTP/2 client connections.
   * `settings` {[Settings Object][]} The initial settings to send to the
     remote peer upon connection.
   * ...: Any [`tls.createServer()`][] options can be provided. For
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -611,7 +611,9 @@ server.listen(80);
   * `noRecvClientMagic` {boolean} (TODO: Add detail)
   * `paddingStrategy` {number} (TODO: Add detail)
   * `peerMaxConcurrentStreams` {number} (TODO: Add detail)
-  * `allowHTTP1` {boolean} Incoming client connections that do not support HTTP/2 will be downgraded to HTTP/1.x when set to `true`. The default value is `false`, which rejects non-HTTP/2 client connections.
+  * `allowHTTP1` {boolean} Incoming client connections that do not support 
+    HTTP/2 will be downgraded to HTTP/1.x when set to `true`. The default value
+    is `false`, which rejects non-HTTP/2 client connections.
   * `settings` {[Settings Object][]} The initial settings to send to the
     remote peer upon connection.
   * ...: Any [`tls.createServer()`][] options can be provided. For
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -181,6 +181,10 @@ class Http2ServerRequest extends Readable {
     return '2.0';
   }
 
+  get socket() {
+    return this.stream.session.socket;
+  }
+
   _read(nread) {
     var stream = this[kStream];
     if (stream) {
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -181,10 +181,6 @@ class Http2ServerRequest extends Readable {
     return '2.0';
   }
 
-  get socket() {
-    return this.stream.session.socket;
-  }
-
   _read(nread) {
     var stream = this[kStream];
     if (stream) {
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1257,7 +1257,8 @@ function connectionListener(socket) {
   }
 
   // TLS ALPN fallback to HTTP/1.1
-  if (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1') {
+  if ((options.allowHTTP1 === undefined || !!options.allowHTTP1 === true) &&
+    (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1')) {
     return httpConnectionListener.call(this, socket);
   }
 
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1257,8 +1257,9 @@ function connectionListener(socket) {
   }
 
   // TLS ALPN fallback to HTTP/1.1
-  if ((options.allowHTTP1 === undefined || !!options.allowHTTP1 === true) &&
-    (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1')) {
+  if (options.allowHTTP1 === true &&
+    (socket.alpnProtocol === false ||
+    socket.alpnProtocol === 'http/1.1')) {
     return httpConnectionListener.call(this, socket);
   }
 
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -13,6 +13,7 @@ const { Duplex } = require('stream');
 const { URL } = require('url');
 const { onServerStream } = require('internal/http2/compat');
 const { utcDate } = require('internal/http');
+const { _connectionListener: httpConnectionListener } = require('http');
 
 const {
   assertIsObject,
@@ -1255,6 +1256,11 @@ function connectionListener(socket) {
     socket.on('timeout', socketOnTimeout);
   }
 
+  // TLS ALPN fallback to HTTP/1.1
+  if (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1') {
+    return httpConnectionListener.call(this, socket);
+  }
+
   socket.on('error', socketOnError);
   socket.on('resume', socketOnResume);
   socket.on('pause', socketOnPause);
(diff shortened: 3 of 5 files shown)


mcollina (commented):
This approach is slightly different to what we were thinking originally, adding an "upgrade" path to an existing https server, so that two different handlers could be installed. However, I like this.

The only problem I am seeing in this is "how do a user know if the current request is HTTP1.1 or HTTP2"?

lib/internal/http2/core.js
```diff
@@ -1303,6 +1304,12 @@ function connectionListener(socket) {
     socket.setTimeout(this.timeout);
     socket.on('timeout', socketOnTimeout);
   }
+
+  // TLS ALPN fallback to HTTP/1.1
+  if (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1') {
+    return httpConnectionListener.call(this, socket);
+  }
```
mcollina: can you add an option to let a user opt-out from upgrading?

sebdeckers:
> @mcollina The only problem I am seeing in this is "how do a user know if the current request is HTTP1.1 or HTTP2"?

[message.httpVersion](https://nodejs.org/api/http.html#http_message_httpversion)
```
request.httpVersion; // "2.0" or "1.1" or "1.0" etc.
```

mcollina:
@sebdeckers can you add it to the test? also with an http2 request, so that we show the difference.

mcollina:
@mikeal you had some good opinions on the design of this step of the API. What do you think?

sebdeckers:
@mcollina I've created an option called `allowHTTP1`. It can be passed to `Http2SecureServer`. The default value is `true`. Setting to `false` prevents the ALPN HTTP/1.1 fallback. This should be documented, perhaps when #102 lands.

Tests cover every combination of https/http2 clients against servers with/without ALPN fallback. Also explicitly reporting the `httpVersion` from server to the client.

jasnell:
fyi... needs a rebase :-)

sebdeckers:
@jasnell Nice, only took me a few minutes this time. Starting to get the hang of `git cherry-pick`. 😎

sebdeckers:
Nits and code issues aside, this implementation hinges on `http._connectionListener`.

#### Con

- It has an underscore prefix and is not documented. This means it is not a public interface?
- `node-spdy` and `node-http2` extend `https.Server`. I don't know if we want to make `Http2SecureServer` extend `https.Server`?

#### Pro

- `require('https')` imports `http._connectionListener` which imports it from `require('_http_server')`. This may imply that `http._connectionListener` is a public interface? At least for core modules?
- Core modules could import `_connectionListener` directly from `require('_http_server')`. Then it stays private, but somewhat couples `http2` and `http` modules.

See also: https://github.com/nodejs/node/issues/2017#issuecomment-114441682

lib/internal/http2/compat.js
```diff
@@ -181,6 +181,10 @@ class Http2ServerRequest extends Readable {
     return '2.0';
   }
 
+  get socket() {
+    return this.stream.session.socket;
+  }
```
mcollina: why? I do not think we should expose this. The socket in the HTTP1 is very different from the socket in HTTP2.

mcollina:
> It has an underscore prefix and is not documented. This means it is not a public interface?

I think overall this uses less of the internals compared to node-spdy and node-http2, so ok for me.

> node-spdy and node-http2 extend https.Server. I don't know if we want to make Http2SecureServer extend https.Server?

IMHO that causes most of the issues in those modules. This is a clean implementation so no.

> require('https') imports http._connectionListener which imports it from require('_http_server'). This may imply that http._connectionListener is a public interface? At least for core modules?

I think we can just import from `_http_server.js`, the function can still be underscored. How about we add a comment there?

> Core modules could import _connectionListener directly from require('_http_server'). Then it stays private, but somewhat couples http2 and http modules.

Coupling is needed if we want to support both http1 and http2.

lib/internal/http2/core.js
```diff
@@ -1255,6 +1256,12 @@ function connectionListener(socket) {
     socket.on('timeout', socketOnTimeout);
   }
 
+  // TLS ALPN fallback to HTTP/1.1
+  if ((options.allowHTTP1 === undefined || !!options.allowHTTP1 === true) &&
+    (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1')) {
+    return httpConnectionListener.call(this, socket);
+  }
```
mcollina: can you put `allowHTTP1` in a variable?

lib/internal/http2/core.js
```diff
@@ -1255,6 +1256,12 @@ function connectionListener(socket) {
     socket.on('timeout', socketOnTimeout);
   }
 
+  // TLS ALPN fallback to HTTP/1.1
+  if ((options.allowHTTP1 === undefined || !!options.allowHTTP1 === true) &&
```
mcollina: why `!!`?

lib/internal/http2/compat.js
sebdeckers: Oops, sorry for sneaking it into this PR. 🙈

In this case I'm using it to report the `alpnProtocol` from the underlying TLS connection. I think there could be other valid use cases. $0.02

If we explicitly *do not* implement this part of the compatibility layer, maybe it should throw or emit a warning?

lib/internal/http2/core.js
sebdeckers: Saw this pattern used elsewhere in Node.js core to coerce to Boolean. Thought it was a convenience to the user. Do you prefer to just do `options.allowHTTP1 === true`?

lib/internal/http2/core.js
sebdeckers: Do you mean a local variable here, for code clarity? Or adding it to the `options` object somewhere else? I did this hacky comparison to avoid setting an extra variable on every connection. Might be a pointless "optimisation."

lib/internal/http2/compat.js
mcollina: > If we explicitly do not implement this part of the compatibility layer, maybe it should throw or emit a warning?

I think we should just throw. The meaning is so different that we can't really provide an equivalent. If you use the socket directly with `require('http')`, very likely your code will not work with `require('http2')`.

> In this case I'm using it to report the alpnProtocol from the underlying TLS connection. I think there could be other valid use cases. $0.02

You need this just for tests?

https://github.com/nodejs/http2/pull/125/files#diff-23822d26e014a01b71d6319c40bef2a6R27

I do not think that is a valid use case.

lib/internal/http2/core.js
mcollina: You are doing two object property lookup here vs just one of the variable.

lib/internal/http2/compat.js
sebdeckers: Okay, fair points. 🤔 Will socket be removed from session at some point? I could change the test case to `request.stream.session.socket.alpnProtocol`.

lib/internal/http2/core.js
mcollina: I think we should not coerce it for every connection. If we need to sanitize this value, we should do it elsewhere.

test/parallel/test-http2-https-fallback.js
```diff
@@ -0,0 +1,130 @@
...
+        }));
+      })
+    );
+  }));
+}
+
+// HTTP/2-only server
+{
+  const server = createSecureServer(
+    { cert, key, allowHTTP1: false },
```
mcollina: I think `allowHTTP1` should be considered `false`  by default, and enable only if it is true.

lib/internal/http2/compat.js
mcollina: Actually, you are right in adding this, as its in the http docs. It does not make much sense to use it for http2, as it will be implementation specific (http, https, http2), and produce different result.

My concern is that, by providing this, we are enabling our users to wreck the internals.
It can stay for now. @jasnell what do you think?

mcollina (approved):
LGTM

sebdeckers:
(Rebased again)

jasnell (approved):

doc/api/http2.md
```diff
@@ -611,6 +611,7 @@ server.listen(80);
   * `noRecvClientMagic` {boolean} (TODO: Add detail)
   * `paddingStrategy` {number} (TODO: Add detail)
   * `peerMaxConcurrentStreams` {number} (TODO: Add detail)
+  * `allowHTTP1` {boolean} Incoming client connections that do not support HTTP/2 will be downgraded to HTTP/1.x when set to `true`. The default value is `false`, which rejects non-HTTP/2 client connections.
```
jasnell: Long line. please wrap at <= 80 chars

mcollina:
Landed as 6232681

mcollina:
My mistake, I force-pushed 5cc2b2a.

PR #127, opened 2017-05-16
http2: allow specifying servername in options

jmuk:
I've noticed that it always use the host name for the SNI parameter, but I think it's better to allow customizing through the options.

##### Checklist

http2


jasnell (approved):
woo! thank you, you just marked an item off my todo list for the week :-)

jasnell:
One thing to think about.. is there anything we need to do on the *server side* to make using SNI easier? (I honestly don't know)

jmuk:
I actually don't know about the server-side SNI parameter, but it looks like to allow customizing the hostname through `addContext` (https://nodejs.org/dist/latest-v6.x/docs/api/tls.html#tls_server_addcontext_hostname_context).

jasnell:
Landed

PR #129, opened 2017-05-16
http2: improve header single value checking

jasnell:
this could be a lot better but it's an improvement over what it was. Will need to play around with the implementation to come up with something that performs better.

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/util.js
+++ b/lib/internal/http2/util.js
@@ -3,6 +3,89 @@
 const binding = process.binding('http2');
 const errors = require('internal/errors');
 
+const {
+  HTTP2_HEADER_STATUS,
+  HTTP2_HEADER_METHOD,
+  HTTP2_HEADER_AUTHORITY,
+  HTTP2_HEADER_SCHEME,
+  HTTP2_HEADER_PATH,
+  HTTP2_HEADER_AGE,
+  HTTP2_HEADER_AUTHORIZATION,
+  HTTP2_HEADER_CONTENT_ENCODING,
+  HTTP2_HEADER_CONTENT_LANGUAGE,
+  HTTP2_HEADER_CONTENT_LENGTH,
+  HTTP2_HEADER_CONTENT_LOCATION,
+  HTTP2_HEADER_CONTENT_MD5,
+  HTTP2_HEADER_CONTENT_RANGE,
+  HTTP2_HEADER_CONTENT_TYPE,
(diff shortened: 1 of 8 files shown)


lib/internal/http2/util.js
```diff
@@ -156,24 +246,25 @@ function mapToHeaders(map) {
...
-          throw new errors.Error('ERR_HTTP2_PSEUDO_HEADERS_SINGLE_VALUE');
+          throw new errors.Error('ERR_HTTP2_HEADER_SINGLE_VALUE', key);
         value = value[0];
       }
-      val = String(value);
-      assertIllegalConnectionSpecificHeader(key, val, mapToHeaders);
-      ret.unshift([key, val]);
+      ret.unshift([key, String(value)]);
     } else {
+      key = String(key).toLowerCase();
```
sebdeckers: Headers must be lower case. Anything upper case is a protocol error. Not sure we should allow it.

lib/internal/http2/util.js
```diff
@@ -145,6 +188,53 @@ function assertIllegalConnectionSpecificHeader(name, value, ctor) {
...
+    case HTTP2_HEADER_AUTHORITY:
+    case HTTP2_HEADER_SCHEME:
+    case HTTP2_HEADER_PATH:
+      return;
+    default:
+      throw new errors.Error('ERR_HTTP2_INVALID_PSEUDOHEADER', name);
+  }
+}
+
+// TODO(jasnell): Find a faster way to do this
```
cjihrig: Would it be faster to load these into a Map and do a lookup here?

sebdeckers (changes requested):

lib/internal/http2/util.js
jasnell: Not quite. Header field names are case insensitive and must only be transmitted in hpack as lower case. Converting to lowercase would be perfectly acceptable.

lib/internal/http2/util.js
jasnell: Possibly. I'm going to be benchmarking it tomorrow

lib/internal/http2/util.js
sebdeckers: Ah, my bad. 😌

mcollina (approved):
LGTM

jasnell:
Landed

PR #130, opened 2017-05-17
http2: request.socket|connection

sebdeckers:
As discussed in: https://github.com/nodejs/http2/pull/125#discussion_r116943863

Adds `request.socket` and `request.connection` for compatibility with documented HTTP APIs:

- > message.socket
  > Added in: v0.3.0
  > <net.Socket>
  > The net.Socket object associated with the connection.
  > https://nodejs.org/api/http.html#http_message_socket
- > Event: 'connection'
  > [...]
  > The socket can also be accessed at request.connection.
  > https://nodejs.org/api/http.html#http_event_connection

Use case: I found that popular packages like [cookies](https://npmjs.org/package/cookies) (NPM: >500k/mo) [make use of this to detect the TLS context](https://github.com/pillarjs/cookies/blob/master/index.js#L87).

##### Checklist

http2

### DIFF
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -227,8 +227,6 @@ process has completed.
 A reference to the [`net.Socket`][] or [`tls.TLSSocket`][] to which this
 `Http2Session` instance is bound.
 
-Directly reading from and writing to the socket is strongly discouraged.
-
 #### http2session.state
 
 * Value: {Object}
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -227,6 +227,8 @@ process has completed.
 A reference to the [`net.Socket`][] or [`tls.TLSSocket`][] to which this
 `Http2Session` instance is bound.
 
+Directly reading from and writing to the socket is strongly discouraged.
+
 #### http2session.state
 
 * Value: {Object}
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -181,6 +181,14 @@ class Http2ServerRequest extends Readable {
     return '2.0';
   }
 
+  get socket() {
+    return this.stream.session.socket;
+  }
+
+  get connection() {
+    return this.socket;
+  }
+
   _read(nread) {
     var stream = this[kStream];
     if (stream) {
--- a/test/parallel/test-http2-compat-serverrequest.js
+++ b/test/parallel/test-http2-compat-serverrequest.js
@@ -3,6 +3,7 @@
 const common = require('../common');
 const assert = require('assert');
 const h2 = require('http2');
+const net = require('net');
 
 // Http2ServerRequest should expose convenience properties
 
@@ -22,8 +23,8 @@ server.listen(0, common.mustCall(function() {
     assert.strictEqual(request.httpVersionMajor, expected.httpVersionMajor);
     assert.strictEqual(request.httpVersionMinor, expected.httpVersionMinor);
 
-    assert.ok(request.socket);
-    assert.ok(request.connection);
+    assert.ok(request.socket instanceof net.Socket);
+    assert.ok(request.connection instanceof net.Socket);
     assert.strictEqual(request.socket, request.connection);
 
     response.stream.on('finish', common.mustCall(function() {
--- a/test/parallel/test-http2-compat-serverrequest.js
+++ b/test/parallel/test-http2-compat-serverrequest.js
@@ -22,6 +22,10 @@ server.listen(0, common.mustCall(function() {
     assert.strictEqual(request.httpVersionMajor, expected.httpVersionMajor);
     assert.strictEqual(request.httpVersionMinor, expected.httpVersionMinor);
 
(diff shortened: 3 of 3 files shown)


mcollina (changes requested):
Can you add a test that verifies how this is used by `cookies`?

sebdeckers:
@mcollina Do you mean just replicate that logic in our test, or to actually pull in the `cookies` package from NPM and use it in a test? (If the latter, what is the recommended way to add 3rd party packages to the node repo?)

mcollina:
We need a line that just does this: https://github.com/pillarjs/cookies/blob/master/index.js#L87

sebdeckers:
Okay, I'm just doing a look around for other examples.

#### finalhandler (NPM: 14M/mo)

> ```js
> req.socket.destroy()
> ```
> https://github.com/pillarjs/finalhandler/blob/master/index.js#L130

#### express (NPM: ∞/mo)

> ```js
> req.connection.encrypted = true;
> ```
> https://github.com/expressjs/express/blob/master/test/req.protocol.js#L41
> ```js
> var proto = this.connection.encrypted
>   ? 'https'
>   : 'http';
> var trust = this.app.get('trust proxy fn');
>
> if (!trust(this.connection.remoteAddress, 0)) {
>   return proto;
> }
> ```
> https://github.com/expressjs/express/blob/master/lib/request.js#L307-L314

mcollina:
https://github.com/pillarjs/finalhandler/blob/master/index.js#L127-L132 cannot be supported as HTTP1 specific. In that specific case for HTTP1, we would have to destroy the stream, not the connection.

`finalhandler` also uses `res._header`  which was not documented in core. The one we should support is https://nodejs.org/api/http.html#http_response_headerssent. I will send a PR to `finalhandler`.

sebdeckers:
#### hapi

```js
    // Setup timeout

    if (this.raw.req.socket &&
        this.route.settings.timeout.socket !== undefined) {

        this.raw.req.socket.setTimeout(this.route.settings.timeout.socket || 0);    // Value can be false or positive
    }
```

https://github.com/hapijs/hapi/blob/master/lib/request.js#L344-L350

mcollina:
I'm convinced. We need this :).

In the test, can you check that whatever `socket`  returns is correct? Just asserting its truthness is not enough.

sebdeckers:
Added a simple `typeof` check which should cover all the public interfaces in use today.

Should I add a test case for HTTPS too? There's not really a difference in core/compat (both are just `Http2Session`). And tls.TLSSocket is a subclass of net.Socket anyway.

mcollina:
that's fine as it is

mcollina (approved):
LGTM

cjihrig (approved):

lib/internal/http2/compat.js
```diff
@@ -181,6 +181,14 @@ class Http2ServerRequest extends Readable {
     return '2.0';
   }
 
+  get socket() {
+    return this.stream.session.socket;
+  }
+
+  get connection() {
+    return this.socket;
```
jasnell, line 189: One point on this: we *really* need to discourage direct access to the socket. With an http/2 connection, bad things can happen if the user does not know what they are doing. The underlying native Http2Session class takes complete ownership of reading and writing to the socket and any direct interactions with the socket by user code could throw the entire connection out of sync. I'd prefer not to expose these at all but if we need to, let's add a strong warning in the docs.

sebdeckers: Okay, will rebase and update the docs that were just landed.

jasnell:
Regarding the hapi setTimeout example... I'm curious why it is necessary actually. In the http/2 case, the `Http2Session` instance itself has a `setTimeout()` and there is a 1-to-1 relationship between the Socket and the Session. Setting the timeout on the `Http2Session` should have the same basic effect, then. In fact, it may be better to accept have `setTimeout()` on `Http2Session` defer to the `setTimeout()` on the socket rather than it being a separate implementation.

sebdeckers:
@jasnell I suspect it stems from wanting to set different timeouts for specific Hapi routes. See: https://github.com/hapijs/hapi/issues/2195 It would only be possible to apply the timeout once the request has been parsed, not at the initial socket connection listener. Just a guess.

jasnell:
Yeah, I get that, but given the multiplexing model, and the fact that individual `Http2Stream` objects have their *own* timeouts that are independent of the `Http2Session` and `Socket`, does that existing model still make sense here?

doc/api/http2.md
```diff
@@ -227,6 +227,8 @@ process has completed.
 A reference to the [`net.Socket`][] or [`tls.TLSSocket`][] to which this
 `Http2Session` instance is bound.
 
+Directly reading from and writing to the socket is strongly discouraged.
```
jasnell: Likely need to take the time to explain why, but this could be a whole section on its own.

mcollina:
Landed as 3904d7c6c142fe492699f7d4d07f4328b1691720

PR #136, opened 2017-05-18
http2: Fix callback in compat

mcollina:
##### Checklist

http2

### DIFF
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -417,7 +417,11 @@ class Http2ServerResponse extends Stream {
   write(chunk, encoding, cb) {
     var stream = this[kStream];
     if (stream === undefined) {
-      cb(new Error('HTTP/2 Stream has been closed'));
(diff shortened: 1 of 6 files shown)


lib/internal/http2/compat.js
```diff
@@ -417,7 +417,11 @@ class Http2ServerResponse extends Stream {
   write(chunk, encoding, cb) {
     var stream = this[kStream];
     if (stream === undefined) {
-      cb(new Error('HTTP/2 Stream has been closed'));
+      var err = new Error('HTTP/2 Stream has been closed');
+      if (cb)
```
sebdeckers, line 421: @mcollina According to docs, all three arguments are optional. Perhaps do something like:

```js
cb = cb || encoding || chunk;
if (typeof cb === 'function') { // ... }
```

Or more correctly:

```js
cb = arguments[arguments.length - 1];
if (typeof cb === 'function') { // ... }
```

sebdeckers (commented):
It looks like this already landed but I have a concern about correctness.

PR #137, opened 2017-05-18
http2: Do not imply a callback to compat res.write

### LINKED ISSUE #135: False assumption that a callback exists?
I was trying to debug the issue #126 so was slowly stepping through in the debugger. So much so that (I presume) something timed out.

I got this error thrown (admittedly before a different error was thrown)

```
internal/http2/compat.js:412
      cb(new Error('HTTP/2 Stream has been closed'));
      ^

TypeError: cb is not a function
    at Http2ServerResponse.write (internal/http2/compat.js:412:7)
    at ReadStream.ondata (_stream_readable.js:574:20)
    at emitOne (events.js:115:13)
    at ReadStream.emit (events.js:210:7)
    at readableAddChunk (_stream_readable.js:195:18)
    at ReadStream.Readable.push (_stream_readable.js:153:10)
    at onread (fs.js:1977:12)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:650:17)

```

mcollina:
Fixes: https://github.com/nodejs/http2/issues/135

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/compat.js
+++ b/lib/internal/http2/compat.js
@@ -416,8 +416,18 @@ class Http2ServerResponse extends Stream {
 
   write(chunk, encoding, cb) {
     var stream = this[kStream];
+
+    if (typeof encoding === 'function') {
+      cb = encoding;
+      encoding = 'utf8';
+    }
+
     if (stream === undefined) {
-      cb(new Error('HTTP/2 Stream has been closed'));
+      var err = new Error('HTTP/2 Stream has been closed');
+      if (cb)
+        cb(err);
+      else
+        this.emit('error', err);
       return;
     }
     var beginSend = this[kBeginSend];
--- a/test/parallel/test-http2-compat-serverresponse-write-no-cb.js
+++ b/test/parallel/test-http2-compat-serverresponse-write-no-cb.js
@@ -1,0 +1,95 @@
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const h2 = require('http2');
+
+// Http2ServerResponse.write does not imply there is a callback
+
+{
+  const server = h2.createServer();
+  server.listen(0, common.mustCall(function() {
+    const port = server.address().port;
+    const url = `http://localhost:${port}`;
+    const client = h2.connect(url, common.mustCall(function() {
+      const headers = {
+        ':path': '/',
+        ':method': 'GET',
+        ':scheme': 'http',
+        ':authority': `localhost:${port}`
+      };
+      const request = client.request(headers);
+      request.end();
+      request.resume();
+    }));
+
+    server.once('request', common.mustCall(function(request, response) {
+      client.destroy();
+      response.stream.session.on('close', common.mustCall(function() {
+        response.on('error', common.mustCall(function(err) {
+          assert.strictEqual(err.message, 'HTTP/2 Stream has been closed');
+        }));
+        response.write('muahaha');
+        server.close();
+      }));
+    }));
+  }));
+}
+
+{
+  const server = h2.createServer();
+  server.listen(0, common.mustCall(function() {
+    const port = server.address().port;
+    const url = `http://localhost:${port}`;
+    const client = h2.connect(url, common.mustCall(function() {
+      const headers = {
+        ':path': '/',
+        ':method': 'get',
+        ':scheme': 'http',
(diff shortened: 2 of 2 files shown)


cjihrig (approved):
LGTM. Was just looking at https://github.com/nodejs/http2/issues/135.

lib/internal/http2/compat.js
```diff
@@ -417,7 +417,11 @@ class Http2ServerResponse extends Stream {
   write(chunk, encoding, cb) {
     var stream = this[kStream];
     if (stream === undefined) {
-      cb(new Error('HTTP/2 Stream has been closed'));
+      var err = new Error('HTTP/2 Stream has been closed');
+      if (cb)
```
jasnell: perhaps `if (typeof cb === 'function')`

cjihrig: So I opened a PR with a change like that in core recently. The pushback was that it introduces silent failure instead of throwing when you try to call a truthy value that isn't actually a function.

jasnell: ugh. ok

sebdeckers: All arguments to `write` are optional. `cb` is only a callback if it's the last argument *and* of type function.

```js
cb = arguments[arguments.length - 1];
if (typeof cb === 'function') { ... }
```

mcollina:
@jasnell @sebdeckers updated.

test/parallel/test-http2-compat-serverresponse-write-no-cb.js
```diff
@@ -0,0 +1,95 @@
...
+        ':scheme': 'http',
+        ':authority': `localhost:${port}`
+      };
+      const request = client.request(headers);
+      request.end();
+      request.resume();
+    }));
+
+    server.once('request', common.mustCall(function(request, response) {
+      response.stream.session.on('close', common.mustCall(function() {
```
mcollina, line 86: @sebdeckers @jasnell do we have a better way to do this?

lib/internal/http2/compat.js
```diff
@@ -416,8 +416,18 @@ class Http2ServerResponse extends Stream {
 
   write(chunk, encoding, cb) {
     var stream = this[kStream];
+
+    if (typeof encoding === 'function') {
```
sebdeckers, line 420: This would fail to invoke the callback for `res.write((err) => ...)`

test/parallel/test-http2-compat-serverresponse-write-no-cb.js
sebdeckers: How about:

```js
function expectedError({message}) {
  strictEqual(message, 'HTTP/2 Stream has been closed')
}
response.on('error', mustCall(expectedError))
response.write()
response.write(mustCall(expectedError))
response.write('stuff', mustCall(expectedError))
response.write('stuff', 'utf8', mustCall(expectedError))
```

lib/internal/http2/compat.js
mcollina: `chunk` is required: https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback

test/parallel/test-http2-compat-serverresponse-write-no-cb.js
mcollina: I meant `response.stream.session.on('close')`. Do we get an event for `'close'` from session  on the stream or req/res?

lib/internal/http2/compat.js
sebdeckers: Oh my bad! Must have had it confused with `.end`

test/parallel/test-http2-compat-serverresponse-write-no-cb.js
sebdeckers: Currently no `close` event is fired on response, afaik. We should implement both `close` and `finish` for compatibility.

lib/internal/http2/compat.js
mcollina: if all of this is ok with you, may I get an LGTM?

sebdeckers (approved):
👍

cjihrig (approved):

mcollina:
Landed in cf8b5da.

PR #144, opened 2017-05-21
http2: Reject incompatible TLS ALPN handshakes

sebdeckers:
1. Only announce `http/1.1` in ALPN when the fallback is allowed.
1. Immediately destroy connections that do not support HTTP/2 when the HTTP/1 fallback is not allowed. Previously these connections would instantiate an Http2Session and linger until they timed out or caused a protocol error.

Note: The Node.js API does not seem to allow rejecting TLS handshakes at the ALPN-level (as described in [RFC 7301 3.2](https://tools.ietf.org/html/rfc7301#section-3.2)). Therefore we do not differentiate between TLS clients that failed to match an ALPN protocol, and clients that simply do not support ALPN. In practice this should make little, if any, difference. Should consider adding strict ALPN negotiation as an option to Node.js core TLS though.

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1619,11 +1619,14 @@ function connectionListener(socket) {
     socket.on('timeout', socketOnTimeout);
   }
 
-  // TLS ALPN fallback to HTTP/1.1
-  if (options.allowHTTP1 === true &&
-    (socket.alpnProtocol === false ||
-    socket.alpnProtocol === 'http/1.1')) {
-    return httpConnectionListener.call(this, socket);
+  if (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1') {
+    if (options.allowHTTP1 === true) {
+      // Fallback to HTTP/1.1
+      return httpConnectionListener.call(this, socket);
+    } else {
+      // Reject non-HTTP/2 client
+      return socket.destroy();
+    }
   }
 
   socket.on('error', socketOnError);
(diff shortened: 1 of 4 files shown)


lib/internal/http2/core.js
```diff
@@ -1425,11 +1425,14 @@ function connectionListener(socket) {
...
-    (socket.alpnProtocol === false ||
-    socket.alpnProtocol === 'http/1.1')) {
-    return httpConnectionListener.call(this, socket);
+  if (socket.alpnProtocol === false || socket.alpnProtocol === 'http/1.1') {
+    if (options.allowHTTP1 === true) {
+      // Fallback to HTTP/1.1
+      return httpConnectionListener.call(this, socket);
+    } else {
+      // Reject non-HTTP/2 client
+      return socket.destroy();
```
sebdeckers: Is `socket.end()` better here?

mcollina: `socket.destroy()`  is actually nice. I'm not sure if we shouldn't even add an `Error` there as first arg.
@jasnell what do you think?

jasnell:
Updated master from nodejs/node master. This will need a rebase

lib/internal/http2/core.js
jasnell: Sorry, just saw this! umm... an error would actually be good here I think.

sebdeckers:
@jasnell rebased 👍

mcollina (approved):
LGTM

sebdeckers:
@mcollina @jasnell As per discussion in the outdated code review I've changed this to emit an error instead of silently destroying the connection. See 	b900d2d. This could be pretty annoying to users, I fear, since must listen for `socketError` events or risk uncaught `error` events being thrown on a the socket (which is only accessible through the `secureConnection` event). Is that by design?

mcollina:
I do not think we should emit an error event. This is a very specific case, and we should be more diligent. So maybe we should use an "unknownProtocol" event passing the socket, and if there are no listeners, we silently drop. If there is a listener (check the return value of emit()), we do nothing and it would be up to the user to decide what they want to do with this.

sebdeckers:
@mcollina That sounds good to me. See ad8bb9d

sebdeckers:
@jasnell Rebased and ready for review. Minor patch but introduces a new event as per above discussion.

mcollina:
Still LGTM :).

jasnell (approved):

jasnell:
Landed!

PR #148, opened 2017-05-31
http2: another round of updates

jasnell:
Starting another round of updates...

##### Checklist

- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines)

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -25,7 +25,8 @@ const {
   getSettings,
   getStreamState,
   mapToHeaders,
-  NghttpError
+  NghttpError,
+  toHeaderObject
 } = require('internal/http2/util');
 
 const {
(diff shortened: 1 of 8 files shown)


lib/internal/http2/core.js
```diff
@@ -1026,7 +1026,6 @@ function streamOnceReady() {
 // on the client side it is used to transmit request data.
 class Http2Stream extends Duplex {
   constructor(session, options) {
-    options.allowHalfOpen = true;
```
mcollina: why?

src/node_http2.cc
```diff
@@ -912,38 +769,33 @@ void Http2Session::OnHeaders(Nghttp2Stream* stream,
...
+      argv[j * 2 + 1] = value_str;
+      headers = item->next;
+      j++;
     }
-    headers = item->next;
-  }
+    if (j > 0) {
+      fn->Call(env()->context(), holder, j * 2, argv).ToLocalChecked();
+    }
+  } while (headers != nullptr);
```
mcollina: can you add some comments explaining how all of this works? It's a bit obscure and it might help in the future.

mcollina (approved):
LGTM with nits

lib/internal/http2/core.js
jasnell: I may end up pulling this particular change out... it appeared to have been causing an issue with some clients but further testing showed some other issues.

jasnell:
Landed!

PR #150, opened 2017-06-07
More updates

jasnell:
More updates... largely fixing and improving settings handling.

##### Checklist

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -26,7 +26,9 @@ const {
   getStreamState,
   mapToHeaders,
   NghttpError,
-  toHeaderObject
+  toHeaderObject,
+  updateOptionsBuffer,
+  updateSettingsBuffer
 } = require('internal/http2/util');
 
 const {
@@ -196,6 +198,7 @@ function onSessionStreamClose(id, code) {
     return;
   _unrefActive(this);
   // Set the rst state for the stream
+  abort(stream);
   stream._state.rst = true;
   stream._state.rstCode = code;
   setImmediate(() => {
@@ -234,10 +237,19 @@ function onSessionRead(nread, buf, handle) {
 
 // Called when the remote peer settings have been updated.
 // Resets the cached settings.
-function onSettings() {
+function onSettings(ack) {
   debug('new settings received');
   _unrefActive(this);
-  this._owner[kRemoteSettings] = undefined;
+  const owner = this._owner;
+  if (ack) {
+    if (owner._state.pendingAck > 0)
+      owner._state.pendingAck--;
+    owner[kLocalSettings] = undefined;
+    owner.emit('localSettings', owner.localSettings);
+  } else {
+    owner[kRemoteSettings] = undefined;
+    owner.emit('remoteSettings', owner.remoteSettings);
+  }
 }
 
 // If the stream exists, an attempt will be made to emit an event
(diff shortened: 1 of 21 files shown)


mcollina (approved):
LGTM

lib/internal/http2/core.js
```diff
@@ -1844,7 +1871,31 @@ function createServer(options, handler) {
...
+                    settings.maxFrameSize,
+                    16384, 2 ** 24 - 1);
+  assertWithinRange('maxConcurrentStreams',
+                    settings.maxConcurrentStreams,
+                    0, 2 ** 31 - 1);
+  assertWithinRange('maxHeaderListSize',
+                    settings.maxHeaderListSize,
+                    0, 2 ** 32 - 1);
+  if (settings.enablePush !== undefined &&
+      typeof settings.enablePush !== 'boolean') {
```
sebdeckers, line 1921: I like how this is written but wonder if it is more efficient to do a strict comparison.

```js
if (settings.enablePush !== undefined &&
    settings.enablePush !== true &&
    settings.enablePush !== false) {
```

Cf [lodash.isBoolean()](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L11396)

lib/internal/http2/core.js
```diff
@@ -1844,7 +1871,31 @@ function createServer(options, handler) {
 // HTTP2-Settings header frame.
 function getPackedSettings(settings) {
   assertIsObject(settings, 'settings');
-  return binding.packSettings(settings || Object.create(null));
+  settings = Object.assign(Object.create(null), settings);
```
sebdeckers: AFAIK we are not mutating `settings` so cloning the object is not necessary. The previous `settings || Object.create(null)` seems preferable.

sebdeckers (commented):
Should `session.settings()` be extended to take an optional callback?

If I understand the spec correctly (6.5.3) there would be one ack per sent settings frame and their order is maintained. So we can use a FIFO queue to track callbacks across multiple pending settings acks.

lib/internal/http2/core.js
jasnell: it could be slightly more efficient but I'm not sure it really buys us much.

lib/internal/http2/core.js
jasnell: ah, right, good point :-)

jasnell:
regarding `settings()` taking a callback, I was stewing over that also. It would be fairly trivial to support. Keep in mind, tho, that there is an initial SETTINGS frame that is sent immediately on connection set up that would get an ack and we really don't have a great way of setting a callback for that one (particularly on the server side). If we added an optional callback, should we *also* still emit `'localSettings'`? (Note that registering the optional callback as a listener for `'localSettings'` won't work since event emitter doesn't have a FIFO capability). @mcollina ... thoughts on this?

jasnell:
Added a bunch more to this PR, including improved options handling and a few more needed tests

test/parallel/test-http2-server-socket-destroy.js
```diff
@@ -45,5 +45,5 @@ server.on('listening', common.mustCall(() => {
   req.on('data', common.mustCall());
 
   client.on('close', common.mustCall());
-  req.end();
+  //req.end();
```
mcollina: why has this left commented out?

jasnell: just missed it, will fix :-)

jasnell:
nit fixed, landed

Issue #153, opened 2017-06-19
Performance Regression: 100% CPU Usage & Memory Leak


sebdeckers:
* **Version**: current master branch (commit: d8b5205a20e06aac27c554678caad08b5b05a4b2)
* **Platform**: macOS Sierra 10.12.5 (16F73)
* **Subsystem**: http2

For as long as a client is connected to an HTTP/2 server (with or without TLS), the node process hogs 100% CPU and memory usage of the process escalates rapidly (10's to 100's of MB/s).

Once the client disconnects the CPU usage returns to idle levels but the memory is not released, not even after forcing GC.

Example test case, a very low traffic client & server:

```js
require('http2').createServer((request, response) => {
  setTimeout(() => response.end('Hello, World!'), 60000)
}).listen(8080)
```

```bash
nghttp -v -s http://localhost:8080/
```

Result, 1 minute later:
<img width="483" alt="screen shot 2017-06-19 at 6 23 47 pm" src="https://user-images.githubusercontent.com/78718/27280957-7c9ff0fc-551c-11e7-8f0b-253599d6713d.png">

On a possibly related note, the h2load rps benchmark figures I'm seeing are an order of magnitude lower than they were a month or so ago.

/cc @jasnell

sebdeckers:
Using `./configure --debug-http2 --debug-nghttp2` (as documented in https://github.com/nodejs/http2/commit/f43597b98f5cd2a82453a3c0d1ebeee22692b305) the node process logs the following line at an incredible rate:

```
stream: adjusting kept idle streams num_idle_streams=5, max=100
stream: adjusting kept idle streams num_idle_streams=5, max=100
stream: adjusting kept idle streams num_idle_streams=5, max=100
stream: adjusting kept idle streams num_idle_streams=5, max=100
[...]
```

🕵️

sebdeckers:
I'm unable to compile anything before the 18785c855018e0319820e3daba5729fd1ab76913 commit. (Was trying to find an older version that did not show the problem, which this commit does still have.)

```
In file included from ../src/node_http2.cc:3:
../src/node_http2.h:558:14: error: calling a protected destructor of class 'node::WriteWrap'
    ::delete req;
             ^
../src/stream_base.h:116:3: note: protected destructor can only be used to destroy a base class subobject
  ~WriteWrap() {
  ^
[...]

PR #154, opened 2017-06-20
http2: Bandaid for infinite loop & memory leak

sebdeckers:
This guard clause eliminates most (like, 99%) of the memory leaking and CPU usage reported in #153. Still seems like there are other leaks elsewhere, but this at least makes the server usable again.

Disclaimer: I don't know what I'm doing in C++ here. 🐒

##### Checklist

http2

### DIFF
--- a/src/node_http2.cc
+++ b/src/node_http2.cc
@@ -821,7 +821,9 @@ void Http2Session::Send(uv_buf_t* buf, size_t length) {
   HandleScope scope(env()->isolate());
   SessionSendBuffer* req = ContainerOf(&SessionSendBuffer::buffer_, buf);
   uv_buf_t actual = uv_buf_init(buf->base, length);
-  if (stream_->DoWrite(req, &actual, 1, nullptr)) {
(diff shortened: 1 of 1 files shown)


src/node_http2.cc
```diff
@@ -821,7 +821,9 @@ void Http2Session::Send(uv_buf_t* buf, size_t length) {
   HandleScope scope(env()->isolate());
   SessionSendBuffer* req = ContainerOf(&SessionSendBuffer::buffer_, buf);
   uv_buf_t actual = uv_buf_init(buf->base, length);
-  if (stream_->DoWrite(req, &actual, 1, nullptr)) {
+  if (length == 0) {
```
jasnell, line 824: Yep, this is where I was going to be looking later on today :-) ... Essentially, after the most recent refactor, the code is a bit too eager to try sending pending data even where there's none pending. I'm going to be taking some time today resolving that but this is a great start.

jasnell:
Issue fixed in #155!

PR #163, opened 2017-06-28
http2: fixed response.end() and cleaned up errors

sebdeckers:
- Fixed `response.end()` so that it works when not sending any headers nor data frames. It just resets the stream without any error. Added a test for this. Also handles the optional encoding/chunk better and now passes the callback to `end` (aka `finish` event) instead of `send`.
- Moved all the error messages into the common errors file. Some are tweaked, hope that's not a problem since we're still pre-release.
- Added a test for sending headers after the stream closes. Throws a proper error message instead of just saying stream is undefined. (This was showing up in my app. Scenario: hammering browser refresh leads to a race condition where connection can be closed before a response has been served.)
- Changed `response.write()` to throw error when no callback is supplied, instead of emitting `error` event. This feels more in line with HTTP/1 implementation.
- Fixed linting in unrelated tests.

##### Checklist

http2

### DIFF
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1599,7 +1599,7 @@ class ServerHttp2Stream extends Http2Stream {
         throw new errors.Error('ERR_HTTP2_STATUS_101');
       if (statusCode < 100 || statusCode >= 200) {
         throw new errors.RangeError('ERR_HTTP2_INVALID_INFO_STATUS',
-                                    statusCode);
+                                    headers[HTTP2_HEADER_STATUS]);
       }
     }
 
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1077,6 +1077,8 @@ function onHandleFinish() {
   if (this[kID] === undefined) {
     this.once('ready', onHandleFinish.bind(this));
   } else {
+    if (this[kState].headersSent === false)
+      this.rstWithNoError();
     const handle = session[kHandle];
     if (handle !== undefined) {
       // Shutdown on the next tick of the event loop just in case there is
@@ -1565,7 +1567,7 @@ class ServerHttp2Stream extends Http2Stream {
       default:
         if (ret < 0) {
           err = new NghttpError(ret);
-          process.nextTick(this.emit('error', err));
+          process.nextTick(() => this.emit('error', err));
           return;
         }
         state.headersSent = true;
@@ -1595,8 +1597,10 @@ class ServerHttp2Stream extends Http2Stream {
       const statusCode = headers[HTTP2_HEADER_STATUS] |= 0;
       if (statusCode === HTTP_STATUS_SWITCHING_PROTOCOLS)
         throw new errors.Error('ERR_HTTP2_STATUS_101');
-      if (statusCode < 100 || statusCode >= 200)
-        throw new errors.RangeError('ERR_HTTP2_INVALID_INFO_STATUS');
+      if (statusCode < 100 || statusCode >= 200) {
+        throw new errors.RangeError('ERR_HTTP2_INVALID_INFO_STATUS',
+                                    statusCode);
+      }
     }
 
     _unrefActive(this);
--- a/lib/internal/http2/core.js
+++ b/lib/internal/http2/core.js
@@ -1077,8 +1077,6 @@ function onHandleFinish() {
   if (this[kID] === undefined) {
     this.once('ready', onHandleFinish.bind(this));
   } else {
-    if (this[kState].headersSent === false)
-      this.rstWithNoError();
     const handle = session[kHandle];
     if (handle !== undefined) {
       // Shutdown on the next tick of the event loop just in case there is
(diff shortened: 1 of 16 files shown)


mcollina (approved):
LGTM

lib/internal/http2/core.js
```diff
@@ -1077,6 +1077,8 @@ function onHandleFinish() {
   if (this[kID] === undefined) {
     this.once('ready', onHandleFinish.bind(this));
   } else {
+    if (this[kState].headersSent === false)
+      this.rstWithNoError();
```
jasnell: hmm... I don't know if we want to do this. If you look at my latest update, `stream.end()` is called automatically when the request is a `HEAD` request because we know there won't be any payload data sent. This way the user will get a write after end error if they try to write payload data for a `HEAD` request. I'm also catching it in `respond()` so that the `options.endStream` flag is set ensuring that no body can be sent at all.

jasnell: @mcollina ... what do you think?

lib/internal/http2/core.js
```diff
@@ -1595,8 +1597,10 @@ class ServerHttp2Stream extends Http2Stream {
       const statusCode = headers[HTTP2_HEADER_STATUS] |= 0;
       if (statusCode === HTTP_STATUS_SWITCHING_PROTOCOLS)
         throw new errors.Error('ERR_HTTP2_STATUS_101');
-      if (statusCode < 100 || statusCode >= 200)
-        throw new errors.RangeError('ERR_HTTP2_INVALID_INFO_STATUS');
+      if (statusCode < 100 || statusCode >= 200) {
+        throw new errors.RangeError('ERR_HTTP2_INVALID_INFO_STATUS',
+                                    statusCode);
```
jasnell: may want to use the pre-transformed `headers[HTTP2_HEADER_STATUS]` value here instead of `statusCode`

test/parallel/test-http2-compat-serverresponse-end.js
```diff
@@ -0,0 +1,42 @@
...
+const { createServer, connect } = require('http2');
+
+// Http2ServerResponse should throw when writing if the stream is closed
+
+const server = createServer();
+server.listen(0, mustCall(() => {
+  const port = server.address().port;
+  server.once('request', mustCall((request, response) => {
+    response.stream.on('finish', mustCall(() => {
+      try {
```
jasnell: why not `assert.throws()` here?

sebdeckers: Hah! Facepalm...  😳 That totally slipped my mind. This whole evening I was feeling uncomfortable about this try/catch mess. Thanks for the reminder.

lib/internal/http2/core.js
sebdeckers: Yea I don't feel great about this either...

Some of the tests send a single whitespace. I thought this was a workaround for this `response.end()` "bug" but maybe it was just a clever way to send some headers and a minimal body.

This was the intended effect:

```js
// Sends HEADERS and DATA(END_STREAM)
require('http2').createServer((request, response) => {
  response.end(' ')
}).listen(8080)

// Sends RST_STREAM(NO_ERROR)
require('http2').createServer((request, response) => {
  response.end() // or: response.end(null)
}).listen(8080)
```

But it could be argued that this is more intuitive:

```js
// Sends HEADERS(END_STREAM)
require('http2').createServer((request, response) => {
  response.end() // or: response.end(null)
}).listen(8080)
```

Sorry to nitpick on such a minor edge-case.

jasnell: What I would expect in that situation is a proper http 200 response with no data. An rst is a bit of a special case that should only be sent when explicitly necessary

sebdeckers:
@jasnell Made it work as per your suggestion. 👌 So, `response.end()` now sends a HEADERS frame and an empty DATA frame with the END_STREAM flag set.

Not really sure why the length=0 DATA frame is sent rather than just setting END_STREAM on the HEADERS frame. Maybe that's just how Nghttp2 does it. Regardless, I believe it's always in the same TCP packet so no real performance impact AFAIK.

Also, for HTTP/1 backwards compatibility, I've normalised the behaviour of `response.flushHeaders`, `response.writeHead`, `response.setHeader`, and added a `response:finish` event. Tests modified to cover their corrected behaviour.

 😅

jasnell:
Regarding the length=0 `DATA` frame... try `stream.respond(headers, { endStream: true })`  :-)

sebdeckers:
@jasnell Thanks, that worked! ☺️ Added a small test for this but it's more or less placebo. The empty `DATA` frames never made it into JS-land anyway, getting filtered out at `Nghttp2Stream::FlushDataChunks`.

jasnell (approved):

jasnell:
Landed!

jasnell:
hmm.. looks like this is actually failing a bit on OSX: https://ci.nodejs.org/job/node-test-commit-osx/10798/nodes=osx1010/console 

Likely due to some intersection with some of the other changes I've made recently. I'll investigate.

jasnell:
May be fixed, may be flaky. Will be keeping an eye on it.

PR #164, opened 2017-06-28
http2: add http2stream.respondWithFile()

jasnell:
Adds a new `http2stream.respondWithFile()` method that allows
sending a file in response without having to go through the
streams mechanism. All of the file i/o happens in the native
layer.

See the docs and test for an example.

##### Checklist

http2

### DIFF
--- a/doc/api/http2.md
+++ b/doc/api/http2.md
@@ -59,6 +59,9 @@ req.end();
 ```
 
 ### Class: Http2Session
+<!-- YAML
+added: REPLACEME
+-->
 
 * Extends: {EventEmitter}
 
@@ -90,10 +93,16 @@ Once a `Socket` has been bound to an `Http2Session`, user code should rely
 solely on the API of the `Http2Session`.
 
 #### Event: 'close'
+<!-- YAML
+added: REPLACEME
+-->
 
 The `'close'` event is emitted once the `Http2Session` has been terminated.
 
 #### Event: 'connect'
+<!-- YAML
+added: REPLACEME
+-->
 
 The `'connect'` event is emitted once the `Http2Session` has been successfully
 connected to the remote peer and communication may begin.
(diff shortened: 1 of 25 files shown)


sebdeckers (commented):
Ooh, very nice! Any performance benchmark numbers of using `nghttp2_submit_response` vs piping streams from JS?

jasnell:
Nothing formal, but testing with h2load shows about a 300% performance improvement.

mcollina:
This is fantastic work, we need this.

Can you please add some more unit tests and checks to cover the following cases:

* [x] sending a file via a push
* [x] the file is removed/altered during send, how are errors handled? does it leak? do we get an event?
* [x] the underlining tcp socket is destroyed, what happens with the errors? does it leak?

Edit: the content length can be specified by the user via the `statCheck` function, but I think we should provide a sane default for that, i.e. sending that header.

jasnell:
I'll add the tests, but in general... when a failure occurs the stream is terminated with an RST stream and the stream is closed. Right now, the only event that would be triggered is the streamClosed event and the normal end/termination events. I've been stewing over what kinds of callbacks to attached to this, if any.

The file descriptor is closed when the stream is closed, which happens after the RST, so we should be good there.

mcollina:
I think we need a callback on this.

jasnell:
Implementation updated... there are now separate `respondWithFD()` and `respondWithFile()` methods. The former takes the `fd` directly without doing a `fs.fstat()` on it. The latter will `fs.open()` then `fs.fstat()` to make sure it is a regular file.

Also, the `statCheck()` function (only available with `respondWithFile()`, can be used to cancel the response by returning `false`... this should make it easier to use with conditional requests.

Also, when a stream is closed and the error code is > 0 (`NGHTTP2_NO_ERROR`), an `'error'` event will also be emitted.

In going through this, I've noticed a few issues with async_hooks. We need to verify that we're properly supporting async_hook context (ping @trevnorris). That can be done in a separate PR.

mcollina (approved):
LGTM

jasnell:
CI: https://ci.nodejs.org/job/node-test-pull-request/8896/

jasnell:
New CI: https://ci.nodejs.org/job/node-test-pull-request/8899/

jasnell:
New CI: https://ci.nodejs.org/job/node-test-pull-request/8907/

jasnell:
Once more, with feeling: https://ci.nodejs.org/job/node-test-pull-request/8908/

jasnell:
failures are unrelated! looks like this is ready to go

mcollina (approved):
LGTM

jasnell:
Landed!

PR #167, opened 2017-07-06
http2: edge case errors, consistency, nits

sebdeckers:
Discovered these while working with everyone's new favourite API, `res.respondWithFile`. 😉

- Cleaned up a stray `console.log` in `test-http2-respond-file.js`. /cc @jasnell 
- Made the compat `res.end(cb)` callback behave the same as http1 (only fires the first time). Also no errors thrown on repeated `res.end()` calls.
- Fixed errors when calling `res.stream.respondWithFile` from the compatibility layer. Seems like a valid use-case (well, I'm using it 🤣).

##### Checklist

http2

### DIFF
--- a/test/parallel/test-http2-compat-serverresponse-end.js
+++ b/test/parallel/test-http2-compat-serverresponse-end.js
@@ -22,7 +22,7 @@ const {
     response.end(mustNotCall());
   }));
   server.listen(0, mustCall(() => {
-    const {port} = server.address();
+    const { port } = server.address();
     const url = `http://localhost:${port}`;
     const client = connect(url, mustCall(() => {
       const headers = {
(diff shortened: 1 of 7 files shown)


sebdeckers:
Pushed a few more changes.

- Achieves the same result without touching `stream._readableState`.
- Additional test case for `HEAD` requests at the compat response layer (was breaking with the earlier patch).
- Expose the internal flag `state.headersSent` as `ServerHttp2Stream.headersSent` to signal whether or not the `.respond()` methods can still be called. Earlier this was tracked separately by `Http2ServerResponse.headersSent` but that went out of sync now that `HEAD` requests and FD's are handled directly by the stream.
- Updated the `var` declarations to `const` and `let` where appropriate. Sorry about the lines changed noise.

sebdeckers:
Added:
- `http2.connect()` documentation shows that `options` and `listener` are optional. Added test cases for this.
- Log output consistency fix to put the session type prefix in square brackets.

test/parallel/test-http2-compat-serverresponse-end.js
```diff
@@ -1,22 +1,28 @@
...
+  // but may be invoked repeatedly without throwing errors.
+  const server = createServer(mustCall((request, response) => {
+    response.end(mustCall(() => {
+      server.close();
     }));
-
+    response.end(mustNotCall());
+  }));
+  server.listen(0, mustCall(() => {
+    const {port} = server.address();
```
jasnell: Minor style nit: `{ port } = ` .... house style is to include spaces around the brackets.

jasnell (approved):

sebdeckers:
@jasnell De-linted. Welcome back! 👋

jasnell:
Landed!

PR #169, opened 2017-07-07
http2: end the stream in statCheck example

sebdeckers:
Just a minor doc fix. I think the example is showing how the send is cancelled and instead a redirect response is issued. The `endStream` option would make sense here, I believe. (Found this when copy-pasting example code like a boss. 😅 Works great so far though!)

##### Checklist

http2 doc


doc/api/http2.md
```diff
@@ -1064,7 +1064,7 @@ const server = http2.createServer();
 server.on('stream', (stream) => {
   function statCheck(stat, headers) {
     // Check the stat here...
-    stream.respond({ ':status': 304 });
+    stream.respond({ ':status': 304 }, { endStream: true });
```
mcollina: why `endStream` here?

sebdeckers: Maybe I misread the example but isn't this code snippet showing a cancellation of the file transfer and instead responding with a Not Modified header frame? The stream remains open unless the endStream flag is set.

mcollina: Does ` stream.respond({ ':status': 304 }).end()` work? This is more idiomatic.

sebdeckers: 1. `.end()` cannot be chained since respond does not return anything.
1. `stream.respond(...); stream.end()` outputs an additional `DATA` frame with the `END_STREAM` flag set. It is cleaner to set the `END_STREAM` flag on the `HEADERS` frame, which is what the `endStream` option on `.respond()` does.

Example: `stream.end()`
```js
  stream.respond({':status': 200})
  stream.end()
```
```
[  0.016] recv (stream_id=13) :status: 200
[  0.016] recv (stream_id=13) date: Fri, 07 Jul 2017 15:48:05 GMT
[  0.016] recv HEADERS frame <length=25, flags=0x04, stream_id=13>
          ; END_HEADERS
          (padlen=0)
          ; First response header
[  0.017] recv DATA frame <length=0, flags=0x01, stream_id=13>
          ; END_STREAM
```

Example: `endStream`

```js
stream.respond({':status': 200}, {endStream: true})
```
```
[  0.013] recv (stream_id=13) :status: 200
[  0.013] recv (stream_id=13) date: Fri, 07 Jul 2017 15:53:29 GMT
[  0.013] recv HEADERS frame <length=25, flags=0x05, stream_id=13>
          ; END_STREAM | END_HEADERS
          (padlen=0)
          ; First response header
```

mcollina: ok, you are right.

jasnell: Side note: I've been thinking about making methods like `respond()` chainable. There are a number of places where we could do it.

jasnell: FYI... the respond method will now automatically default `endStream` to `true` if the status code is `304`. See: https://github.com/nodejs/http2/blob/master/lib/internal/http2/core.js#L1651-L1659

sebdeckers:
No longer required, as explained above.

PR #170, opened 2017-07-07
http2: add test for invalid pseudo headers

yosuke-furukawa:
I add pseudo headers check tests !
Thunks god, it's open source friday!!! https://github.com/blog/2386-contribute-on-open-source-friday

##### Checklist

http2

### DIFF
--- a/test/parallel/test-http2-server-set-invalid-header.js
+++ b/test/parallel/test-http2-server-set-invalid-header.js
@@ -1,0 +1,54 @@
+// Flags: --expose-http2
+'use strict';
+
+const common = require('../common');
+const assert = require('assert');
+const http2 = require('http2');
+const body =
+  'Cannot set HTTP/2 pseudo-headers';
+
+const invalidHeaders = [
+  { key: ':status', value: 200 }, 
+  { key: ':method', value: 'GET' }, 
+  { key: ':path', value: '/' }, 
+  { key: ':authority', value: 'example.com' }, 
+  { key: ':scheme', value: 'http' }
+];
+
+const checkServer = (invalidHeader, value) => {
+  const server = http2.createServer((req, res) => {
+    res.setHeader('foobar', 'baz');
+    res.setHeader('X-POWERED-BY', 'node-test');
+    try {
+      res.setHeader(invalidHeader, value);
+    } catch (e) {
+      res.statusCode = 500;
+      res.end(e.message);
+    }
+  });
+  
+  server.listen(0, common.mustCall(() => {
+    const client = http2.connect(`http://localhost:${server.address().port}`);
+    const headers = { ':path': '/' };
+    const req = client.request(headers);
+    req.on('response', common.mustCall((headers) => {
+      assert.strictEqual(headers[':status'], 500);
+      assert.strictEqual(headers['foobar'], 'baz');
+      assert.strictEqual(headers['x-powered-by'], 'node-test');
+    }));
+  
+    let data = '';
+    req.on('data', (d) => data += d);
+    req.on('end', common.mustCall(() => {
+      assert.strictEqual(body, data);
+      server.close();
+      client.destroy();
+    }));
+    req.end();
+  }));
+  server.on('error', common.mustNotCall());
+};
+
+invalidHeaders.forEach(({ key, value }) => {
+  checkServer(key, value);
+});


test/parallel/test-http2-server-set-invalid-header.js
```diff
@@ -0,0 +1,56 @@
...
+  { key: ':method', value: 'GET' }, 
+  { key: ':path', value: '/' }, 
+  { key: ':authority', value: 'example.com' }, 
+  { key: ':scheme', value: 'http' }
+];
+
+const checkServer = (header, value) => {
+  const server = http2.createServer((req, res) => {
+    try {
+      res.setHeader('foobar', 'baz');
```
sebdeckers: Move these first two `setHeader` calls out of the `try` block. They should never throw.

test/parallel/test-http2-server-set-invalid-header.js
```diff
@@ -0,0 +1,56 @@
...
+      res.setHeader('X-POWERED-BY', 'node-test');
+      // invalid header
```
sebdeckers: Remove this comment and rename the `header` variable as `invalidHeader`.

sebdeckers (commented):
FYI there may be some overlap with these tests:
https://github.com/nodejs/http2/blob/master/test/parallel/test-http2-misused-pseudoheaders.js
Is it best to separate these test cases or combine them?

test/parallel/test-http2-server-set-invalid-header.js
```diff
@@ -0,0 +1,56 @@
...
+    const headers = { ':path': '/' };
+    const req = client.request(headers);
+    req.on('response', common.mustCall((headers) => {
+      assert.strictEqual(headers[':status'], 500);
+      assert.strictEqual(headers['foobar'], 'baz');
+      assert.strictEqual(headers['x-powered-by'], 'node-test');
+    }));
+  
+    let data = '';
+    req.on('data', (d) => data += d);
```
mcollina, line 41: this should require a `setEncoding('utf8')` as _it should_ come out as `Buffer` here. I assume this test is passing, so maybe there is something else going on.

cc @sebdeckers @jasnell

mcollina (commented):
@sebdeckers I think it's better to combine the same test into one, as we are essentially testing the same behavior.

test/parallel/test-http2-server-set-invalid-header.js
sebdeckers: Tests are passing. Well spotted, this is weird. Here's what I'm seeing just from the repl.

```js
new stream.Duplex()._readableState.defaultEncoding
// -> 'utf8'
```

mcollina: that should not be. We should be moving buffers.

sebdeckers: I guess this is a thing now...? https://github.com/nodejs/node/blob/85dacd63f0938cf135d49d393ceb9e85b4b1fac5/lib/_stream_readable.js#L111

mcollina: https://github.com/nodejs/node/commit/d5158574c6cb69204d155b6ea0ee4c07ab07ddd6 is very old. So, no there should be something else into play here.

sebdeckers: My bad, I read that all wrong. (deleted previous confused messages)

sebdeckers: @mcollina String concatenation => type coercion?

```js
'foo' + new Buffer('bar')
// 'foobar'
```

mcollina: oh that's fine then.

sebdeckers: Just checked and `d` is in fact a `Buffer`. We're still emitting buffers, not strings, from the `data` event. 😅

addaleax: @sebdeckers The request from @mcollina is the right way to handle this, that’s what `setEncoding` is there for. In your example you are lucky it works, but if the chunks are e.g. `Buffer([0x46, 0xc3])` and `Buffer([0x96, 0x4f])`, the result is `F��O` (edit: and not `FÖO`, as it should be).

sebdeckers: @addaleax Yep, absolutely agreed. I'm just saying there is no deeper bug; nothing else is broken in the codebase. Apologies for adding my own confusion to this discussion. 😖

@yosuke-furukawa Can you please add `res.setEncoding('utf8')` here?

jasnell:
Ping @yosuke-furukawa ... this needs to be rebased :-)

yosuke-furukawa:
sorry for late response, I just fixed. but if it is too late to merge, feel free to ignore this. :)
 I would be better to catch up http2 more.

PR #178, opened 2017-07-12
http2: test cases and other tweaks

jasnell:
Bunch of test cases ported from existing http tests, and a few changes to support them.

Note: this changes the way `mapToHeader` errors are reported on the client. In all cases, `mapToHeader` errors will be emitted in an `'error'` event on the client side. On the server side, most `mapToHeader` errors are thrown synchronously with the notable exception of `mapToHeaders` errors thrown within the `fetchTrailers` event.

Also adds `http2.getUnpackedSettings()` as a reverse to `http2.getPackedSettings()`

##### Checklist

http2


lib/internal/http2/core.js
```diff
@@ -214,18 +222,14 @@ function onSessionTrailers(id) {
...
-  try {
-    const trailers = Object.create(null);
-    stream.emit('fetchTrailers', trailers);
-    return mapToHeaders(trailers, assertValidPseudoHeaderTrailer);
-  } catch (err) {
-    process.nextTick(() => stream.emit('error', err));
+  const trailers = Object.create(null);
+  stream.emit('fetchTrailers', trailers);
+  const headersList = mapToHeaders(trailers, assertValidPseudoHeaderTrailer);
+  if (headersList instanceof Error) {
```
mcollina: instanceof can be slow. Given that this will likely ship with a turbofan-enabled released, can we just throw?

lib/internal/http2/core.js
```diff
@@ -1589,8 +1612,14 @@ class ServerHttp2Stream extends Http2Stream {
       options.endStream = true;
     }
 
+    const headersList = mapToHeaders(headers);
+    if (headersList instanceof Error) {
+      // An error occurred!
+      throw headersList;
+    }
```
mcollina: ditto

lib/internal/http2/core.js
```diff
@@ -2145,7 +2190,7 @@ function clientSessionOnError(error) {
     return;
   this.destroy();
   this.removeListener('error', clientSocketOnError);
-  this.emit('error', error);
+  this.removeListener('error', clientSessionOnError);
```
mcollina, line 2215: what happens if an error is thrown afterwards?

test/parallel/test-http2-compat-serverresponse-statuscode.js
```diff
@@ -21,8 +21,7 @@ server.listen(0, common.mustCall(function() {
     };
     const fakeStatusCodes = {
       tooLow: 99,
-      tooHigh: 1000,
-      backwardsCompatibility: 999
+      tooHigh: 1000
```
sebdeckers: Change `tooHigh` to 600

test/parallel/test-http2-status-code-invalid.js
```diff
@@ -0,0 +1,40 @@
...
+  return common.expectsError({
+    code: 'ERR_HTTP2_STATUS_INVALID',
+    type: RangeError,
+    message: `Invalid status code: ${code}`
+  });
+}
+
+server.on('stream', common.mustCall((stream) => {
+
+  // Anything lower than 100 and greater than 699 is rejected
```
sebdeckers: Should this be `599`?

sebdeckers (commented):
LGTM

BTW does #167 help at all here? It also tries to address issues with write callbacks in the compat layer.

lib/internal/http2/core.js
jasnell: mapToHeaders might return an Array or an Error (yes, I know polymorphic returns are awful). I'm thinking that maybe we can do a less expensive check here.

lib/internal/http2/core.js
jasnell: because of the `this.destroy()` above, that should be fine. Nothing that we need to catch should throw after this point.

test/parallel/test-http2-status-code-invalid.js
jasnell: heh... yes, fingers hit the wrong button :-)

jasnell:
@sebdeckers ... yep, I'm going to be taking a look at #167 later on today and will get that landed before this one.

jasnell:
@mcollina @sebdeckers updated!

lib/internal/http2/util.js
```diff
@@ -50,6 +51,9 @@ const {
   HTTP2_METHOD_HEAD
 } = binding.constants;
 
+// This set is defined strictly by the HTTP/2 specification. Only
+// :-prefixed headers defined by that specification may be added to
```
sebdeckers, line 55: 😛 can't unsee `:-p`

test/parallel/test-http2-client-unescaped-path.js
```diff
@@ -0,0 +1,37 @@
...
+server.on('stream', common.mustNotCall());
+
+const count = 32;
+
+server.listen(0, common.mustCall(() => {
+  const client = http2.connect(`http://localhost:${server.address().port}`);
+
+  let remaining = count;
+  function maybeClose() {
+    if (--remaining === 0) {
```
sebdeckers, line 18: @jasnell OT but I wish `common` provided some utilities like a countdown.

sebdeckers (approved):
LGTM

Thanks for cleaning up `test-http2-compat-serverresponse-createpushresponse.js` 🤝

test/parallel/test-http2-client-unescaped-path.js
jasnell: It's not out of the question :-)

jasnell:
Landed!

